Issue #5, opened 2015-07-14
'groupId' not always pulling from parent pom


roytruelove:
(with version 2.2)

Getting an NPE with the following (pretty standard) scenario using `versions:set`

In my remote repo (e.g. Artifactory) I've got a global parent pom (com.myco:global-parent-pom:pom.1.0-SNAPSHOT)

I then have a project:

```
+---myProject
    +--- pom.xml
    +--- moduleA
         +--- pom.xml
         +--- (etc)
    +--- moduleB
         +--- pom.xml
         +--- (etc)
```

myProject's parent pom references the 'global' parent pom as its parent:

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

        <artifactId>my-project-pom</artifactId>
        <version>1.0.0-SNAPSHOT</version>
        <modelVersion>4.0.0</modelVersion>
        <packaging>pom</packaging>

        <parent>
                <groupId>com.myco</groupId>
                <artifactId>global-parent-pom</artifactId>
                <version>1.0-SNAPSHOT</version>
        </parent>

        <modules>
                <module>moduleA</module>
                <module>moduleB</module>
        </modules>

</project>
```
:sparkles: Important - notice that I am *not* explicitly specifying a groupId for this pom since it should inherit from the parent (com.myco) :sparkles: 

moduleA and moduleB use myProject's pom as their parent.  That's the setup - very standard.  

When I run `versions:set` at the myProject level I get the following NPE stacktrace:

```
[DEBUG] Applying change null:my-project-pom:1.0.0-SNAPSHOT -> 1.0.1-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] moduleA ................................... FAILURE [  8.936 s]
[INFO] moduleB ......................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 9.818 s
[INFO] Finished at: 2015-07-14T12:35:50-04:00
[INFO] Final Memory: 13M/156M
[INFO] ------------------------------------------------------------------------
[...]

PR #6, opened 2015-07-16
'groupId' not always pulling from parent pom

tomerc:


khmarbaise:
I assume that should fix issue #5? A reference to the issue in the log message would be furthermore a test would be very good...

tomerc:
Added testcase

khmarbaise:
Can you please squash your commits...

tomerc:
Done

khmarbaise:
Ok. Integrated. Thanks.

PR #18, opened 2015-08-17
new goal: use-dep-version

DanArcari:
Added new goal (“use-dep-version”), to allow specification of a
specific version for a specific dependency. This is something we have
been using internally to manage versions for a very large (100+ module)
project where we can’t always use the most up-to-date version of a
particular third party component.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDependencyUpdaterMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDependencyUpdaterMojo.java
@@ -343,7 +343,14 @@ public abstract class AbstractVersionsDependencyUpdaterMojo
         }
 
         return result;
-
+    }
+    
+    /**
+     * Indicates whether any includes were specified via the 'includes' or 'includesList' options.
+     * @return true if includes were specified, false otherwise.
+     */
+    protected boolean hasIncludes() {
+    	return includes != null || includesList != null;
     }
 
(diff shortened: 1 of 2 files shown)


DanArcari:
As mentioned in the commit note, this is a goal I've created for internal use that I'd like to integrate with the public version in order to avoid having to maintain a local version. Hopefully others might find it useful as well.

glianeric:
I have a scenario similar to yours, where I basically want to update a dependency from:

2.2.0-SNAPSHOT to 2.2.0.

The dependency version is stored in a property.  Will your new goal process the properties as well?

DanArcari:
@glianeric The goal I added is basically a wrapper around the base plugin functionality, so if the plugin can change versions inside of a property, then the new goal should as well. It's not something I tested, however.

quantoid:
The comments here say the goal is `use-dev-version` but the code says `use-specific-version` with `--depVersion`.

Also this new goal does not appear in the official documentation.

oreporan:
@khmarbaise - why was this linked to issue #39 ? This takes care of update-dep but not of set-property to an older versions (including validating its existence)

oreporan:
I made a pull request to handle updating properties as asked by @glianeric 
pull request #124

Marc-Antoine:
It seems not work anymore

Marc-Antoine:
It seem not work where property is used to set the version. To change the property value I can use versions:update-property but I have no way to know if it is a property or not and the name of the property

PR #27, opened 2015-09-30
fix versions:set when used with an alternate POM

boillodmanuel:
When using an alternate pom with `mvn -f alternate-pom.xml versions:set -DnewVersion=0.0.1`, submodule versions are updated, but not the parent pom version.

This fix works with an alternate pom and with a classic `pom.xml`


khmarbaise:
Based on the integration test this change will break the listed integration tests. Take a look into my comment of the revert.

boillodmanuel:
May be already fixed by  #57 ?

boillodmanuel:
May be already fixed by  #58?

PR #49, opened 2016-01-20
fix for dead url detected by linkcheck maven plugin during others sit…

romani:
…e generation

example of problem detected by linkcheck plugin https://codeship.com/projects/124310/builds/11383081 scroll till the end of log


romani:
is there any problem with such update ?

PR #84, opened 2016-06-13
major,minor and incremental flags for versions in properties

pbromley:
This is something we are missing since a lot of our versions are contained in properties

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/DisplayPropertyUpdatesMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/DisplayPropertyUpdatesMojo.java
@@ -85,6 +85,30 @@ public class DisplayPropertyUpdatesMojo
      */
     private Boolean autoLinkItems;
 
+    /**
+     * Whether to allow the major version number to be changed.
+     *
+     * @parameter property="allowMajorUpdates" default-value="true"
+     * @since 1.2
+     */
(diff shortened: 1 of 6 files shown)


src/main/java/org/codehaus/mojo/versions/DisplayPropertyUpdatesMojo.java
```diff
@@ -85,6 +85,30 @@
      */
     private Boolean autoLinkItems;
 
+    /**
+     * Whether to allow the major version number to be changed.
+     *
+     * @parameter property="allowMajorUpdates" default-value="true"
+     * @since 1.2
+     */
```
khmarbaise: If this an addition than you should correct the @since and give the correct version number (2.3 ?)

khmarbaise:
It would be helpful to squash the commits...

pbromley:
Agreed on the @since and the squashing of the commits - i'll fix that

I started with a TDD approach but found there were no tests covering the same functionality from the use-latest-versions mojo - no excuse in the quest for quality I know.

I'll look at adding some tests if i get time

khmarbaise:
If the existing functionality is not covered it would be great getting some tests for it..if you like..if you find areas where this is true ;-)

tomasbjerre:
Any chance of getting this fixed and merged any time soon?

PR #86, opened 2016-06-16
UseReleases: now with allowRangeMatching and failIfNotReplaced

talksinmath:
In order to use this plugin we had to enhance the behaviour of the `UseReleasesMojo`.

Before, if the version that should be replaced was `1.0-SNAPSHOT`, it would only look for a matching release with the version `1.0`.

Due to our continous-integration pipeline we needed the functionality that `1.0-SNAPSHOT` should be replaced with the highest matching version `1.0.x.y` (e.g. `1.0.1.100`).

This pull request implements this behaviour, it can be activated with the property `allowRangeMatching`. Also, if no matching snapshot is found, the property `failIfNotReplaced` causes the build to immediately fail.

Both properties default to `false` so that the default behaviour is not changed.

We considered using `force-releases`, but that had the effect, that `1.1-SNAPSHOT` would be resolved to `1.2` if such a version exists. We would like the upper-bound of the range to be respected (i.e. nothing above `1.1.x.y`).

We're looking forward to your comments and hope that this functionality is useful to others.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/UseReleasesMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/UseReleasesMojo.java
@@ -21,6 +21,7 @@ package org.codehaus.mojo.versions;
 
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.metadata.ArtifactMetadataRetrievalException;
+import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.model.Dependency;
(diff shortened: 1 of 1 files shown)


khmarbaise:
The PR looks good so far only the absence of one or more tests can be improved ?

talksinmath:
I've added a test and improved the behaviour. Unfortunately, I'll have to close this PR and open another one since I got my branches mixed up.

talksinmath:
Please check here: #87

PR #91, opened 2016-06-28
versions:set - add "removeSnapshot" property (squashed)

bpindelski:
This PR supersedes https://github.com/mojohaus/versions-maven-plugin/pull/52. I took the liberty of squashing @designhaiku's commits into one, as per @tchemit's request.


khmarbaise:
The problem i have seeing what is really changed based on the format change...can you fix that?

bpindelski:
@khmarbaise Certainly. I was trying to not to interfere with the original commits, but the formatting changes could be dropped from this PR.

Issue #101, opened 2016-08-26
versions:set nextSnapshot is only working, when last number before -SNAPSHOT is zero-filled three digit.


rherschke:
The code, used in SetMojo when running with `nextSnapshot = true` is only working, when the version of the project is in the format `x.y.zzz-SNAPSHOT`.

It assumes, that the "version" is a three-digit, zero filled Integer number (due to ominous magic number 3 in `substring(length-3)`), which leads to a NumberFormatException when running this with "normal" maven version patterns like 1.2.3-SNAPSHOT or even 1.2-SNAPSHOT and therefore this option is useless in most cases.

Please correct it by supporting all version formats.

Thx
Robert

lion7:
I proposed a fix in https://github.com/mojohaus/versions-maven-plugin/pull/105

OlaiSolheim:
I'm rooting for this. :)

Frettman:
I registered at GitHub just to say how silly it is to assume that the last version segment has three digits. Not even the **versions-maven-plugin** follows that "convention". I hope this gets fixed soon, because I was happy to find the new feature...until it failed with a NumberFormatException :)

timpeeters:
Great news that this has been fixed. Any idea when we might expect a new release of the plugin containing this fix?

Thanks in advance!

foal:
Release?

PR #105, opened 2016-10-07
Fix for issue #101: nextSnapshot requires the last part of the version to be exactly 3 digits

lion7:
See https://github.com/mojohaus/versions-maven-plugin/issues/101 for details on the issue.
I added a few tests to prove that removeSnapshot and nextSnapshot work as intended.


lion7:
Closed because https://github.com/mojohaus/versions-maven-plugin/issues/101 is fixed in https://github.com/mojohaus/versions-maven-plugin/commit/eebc0eff1540ad0fb502247f99cb15648fc1d4ff

PR #108, opened 2016-10-11
Execute <use-releases> goal in the corresponding example instead of wrong <unlock-snapshot> goal

VoldemarLeGrand:
Example of `use-releases` goal in documentation by mistake execute `unlock-snapshot` goal. 
This patch correct the command to execute in documentation.


khmarbaise:
Unfortunately fixed by #118 many thanks for your help...

PR #119, opened 2016-11-28
Add an option to SetMojo to set some properties to newVersion

ppalaga:
@khmarbaise could you please review?

This is primarily motivated by projects like [1] which manage their own modules
using a version property. In the current versions plugin there is no
simple way to set this property by a single versions:set invocation. 

After this change it will be possible to configure the versions-maven-plugin to set any desired property  [2] to the `newVersion` automagically, e.g. when releasing.

[1] https://github.com/pietermartin/sqlg/blob/master/pom.xml#L191-L205 note that `sqlg.version` is supposed to be kept in sync with `project.version`  https://github.com/pietermartin/sqlg/blob/master/pom.xml#L34 https://github.com/pietermartin/sqlg/blob/master/pom.xml#L16

[2] https://github.com/mojohaus/versions-maven-plugin/commit/bdee9e04c777e50f76c1ecba359ab9106850e1a1#diff-13791dad6080eb3527564192fbbc5f5eR113

### DIFF
--- a/src/it/it-set-010/invoker.properties
+++ b/src/it/it-set-010/invoker.properties
@@ -1,0 +1,3 @@
+invoker.goals=${project.groupId}:${project.artifactId}:${project.version}:set -DnewVersion=1.2.1-SNAPSHOT -DsetProperties=myProperty1
+invoker.nonRecursive=true
+invoker.buildResult=success
--- a/src/it/it-set-010/module-a1/pom.xml
+++ b/src/it/it-set-010/module-a1/pom.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <parent>
+    <groupId>localdomain.localhost</groupId>
+    <artifactId>project-a</artifactId>
+    <version>1.2.1-SNAPSHOT</version>
+  </parent>
+
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>localdomain.localhost</groupId>
+  <artifactId>module-a1</artifactId>
+  <packaging>pom</packaging>
+  <version>2.0.7-SNAPSHOT</version>
+
+</project>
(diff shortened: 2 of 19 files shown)


ppalaga:
@khmarbaise is there anything missing in this PR so that I can get a review?

oreporan:
👍

ppalaga:
@oreporan thanks for upvoting. 

Looks like your https://github.com/mojohaus/versions-maven-plugin/pull/124 tries to solve the same kind of problem via another versions plugin mojo. I must say I went your path first too, but then I realized having the ability to set the property to `newVersion` in the `set` mojo is way more practical and will seamlessly work during releases too. 

That said, I do not think this PR is really a duplicate of your #124 . Having `forceNewVersion` option in `update-property` will help people who do not/cannot use `set` mojo for some reason.

khmarbaise:
Maybe I miss a thing here but why could `update-property` not being used for this scenario ?

ppalaga:
> why could update-property not being used for this scenario ?

First, what I need cannot be done with `update-property`, because `update-property` only updates to the latest versions and it ignores `newVersion`. BTW, this is what #124 attempts to address.

Second, for my use case, it is way more useful to be able to set the property through the `set` mojo, because I want to do it during releases. I doubt it is possible to set up the release plugin to invoke two distinct versions plugin mojos. I tried it but I failed. Having the option to set a property to newVersion would make the release-plugin set-versions-plugin duo much easier to use.

khmarbaise:
The SetMojo is intended for setting version of pom and not for setting properties. That's why update-properties, update-property exists. But it sounds like it makes more and more sense to introduce some other goals related to properties like `properties-next-release`, `properties-latest-versions` (need to think about that in deepth) etc. Apart from that mixing different concerns into a single goal looks bad to me..Furthermore calling different goals during the release is possible which can be defined by using `goals` parameter of the [maven-release-plugin](http://maven.apache.org/maven-release/maven-release-plugin/perform-mojo.html#goals). Furthermore can you give me concrete examples where you need to run versions-maven-plugin during the release process from maven-release-plugin ?

ppalaga:
> The SetMojo is intended for setting version of pom and not for setting properties. That's why update-properties, update-property exists. 

I'd dare to re-state that `update-properties` and `update-property` ignore `newVersion` and hence cannot be used to set a property to an arbitrary version. Even if this is addressed (e.g. by merging #124 ), a solution with `update-property` would mean much more lines of configuration than a solution based on the present PR. 

> But it sounds like it makes more and more sense to introduce some other goals related to properties like properties-next-release, properties-latest-versions (need to think about that in deepth) etc. 

I am not asking for any of those.

> Apart from that mixing different concerns into a single goal looks bad to me.

A project may choose to want to keep some property in sync with the current project's version. Setting `versions:set` to ensure that sounds much more natural and safer to me than figure out a way to enforce that `update-property` is always run after `set`. That would be error prone. 

> Furthermore calling different goals during the release is possible which can be defined by using goals parameter of the maven-release-plugin. 

This will perhaps be possible after merging #124 with a lot of xml config that I am not looking forward for.

> Furthermore can you give me concrete examples where you need to run versions-maven-plugin during the release process from maven-release-plugin ?

E.g. here in `sqlg` they need to keep `sqlg.version` property in sync with the current project's version. https://github.com/pietermartin/sqlg/blob/master/pom.xml#L34

Issue #98, opened 2016-07-29
Effect of update-properties has changed in v2.3


king-aegon:
Previously (in v2.2), when running mvn version:update-properties, properties which were SNAPSHOT versions would not be updated to a released version unless a released version of equal version was available.

eg. running mvn versions:update-properties for property
`<myapp.version>4.7.0-SNAPSHOT</myapp.version>`
this property would only be updated if 4.7.0 had been released

Now (in v2.3) this has changed.  Running mvn version:update-properties is reverting all SNAPSHOT properties to the most recently released version regardless of the SNAPSHOT's version number.

In the example above, the myapp.version property is being changed to 4.6.2.

Is this change intentional?  The previous operation was much more suited to our needs.

khmarbaise:
Have you tried to ue `-DallowSnapshots=false`? And yes there was a change #33

king-aegon:
Using mvn versions:update-properties -DallowSnapshots=false on the example above results in the same thing -> it still updates 4.7.0-SNAPSHOT to 4.6.2

I only want the SNAPSHOT property to be updated once 4.7.0 is released.  Until then it should remain as 4.7.0-SNAPSHOT

hSpille:
During development cycle it is not helpful when Mr. Jenkins is always reverting to an older release version of our module. So, we really want this "feature" to be reverted - or at least possible to be turned-off by arguments. 

Anyway - thx for developing versions-plugin: Nice work!!!11eleven

binkley:
I'm getting related non-snapshot behavior as well.  Toggling between versions plugin 2.2 and 2.3, `update-properties` *downgrades* some plugins.  Affected plugins are those I manually bumped to latest version from what maven 3.3.9 specifies.  Version 2.2 does not have this behavior.

I refer to all plugin versions via properties.

glianeric:
> Using mvn versions:update-properties -DallowSnapshots=false on the example above results in the same thing -> it still updates 4.7.0-SNAPSHOT to 4.6.2

Please use -DallowSnapshots=true

> I'm getting related non-snapshot behavior as well. Toggling between versions plugin 2.2 and 2.3, update-properties downgrades some plugins. Affected plugins are those I manually bumped to latest version from what maven 3.3.9 specifies. Version 2.2 does not have this behavior.

That's strange - the change #33 was a one-liner IIRC.

glianeric:
Here's the change:

[...]

PR #124, opened 2016-12-25
added boolean forceNewVersion to update-property goal

oreporan:
we update our pom depedencies based on properties and need the abilityn to roll back to older versions automatically based on testing. Currently the update-property goal only updates to the latest versions (effectively ignoring the newVersion property if it isn't the latest)

The fix is a simple if/else on a new boolean property `-DforceNewVersion=true` 
If this is set to true (default false), then the code runs over all the versions of the depedency linked to this property, and as long as this version exists, it is set. If the version doesn't exist at all, this leaves the property unchanged (like the original code)


oreporan:
possible dupliate of #119

sregger:
@oreporan did this change get released? I'm looking to start using

> mvn versions:update-property

with

> -DnewVersion

and like you I'd need to support rollbacks.

oreporan:
@sregger - I have no idea. It appears it was never merged..

sregger:
Pity - thanks for the update @oreporan

khmarbaise:
Hi, currently I'm working on a larger number of changes...at the moment I see that this PR has conflicts in it..furthermore I don't see tests for this new behaviour...It would also helpful having a squashed commit which contains the whole change in a single commit ...

sregger:
@oreporan and @khmarbaise 

I was too quick to hop on this bandwagon. Turns out that -DnewVersion supports rollbacks. But it has bad error detection for arguments.

I was using the command like so

> mvn versions:update-property -DnewVersion=1.0.123 -Dproperty=my-property.version

which **always** upgrades to the latest. I did not notice this as 1.0.123 was my latest. But I did notice I could not downgrade once I'd set 1.0.123.

However this command is incorrect. The correct command is

> mvn versions:update-property -DnewVersion=[1.0.123] -Dproperty=my-property.version

You need to use the range when setting a newVersion. e.g. [1.0.123] versus 1.0.123

Now I'm able to rollback to old versions.

sregger:
Further info

If the property name is specified incorrectly unexpected behavior is encountered.

e.g.

> -Dproperty=my-version instead of my.version

This causes no warning or error. But -DnewVersion is ignored and instead the **latest** version is set. 

> -Dproperty=my-version instead of my.version

In the above the dot (.) was replaced with a dash (-).

This caused no property matches and hence no changes.

It seems maven checks if any property is contained within the provided -DnewVersion

khmarbaise:
The handling for downgrade a version to a previous version has been done with #98  
```
mvn versions:update-property -Dproperty=testng.version -DnewVersion=[5.6] -DallowDowngrade=true
```
this will allow downgrade of this property to version `5.6` ...
Apart from that if you don't see something in the log like:
```
[INFO] --- versions-maven-plugin:2.5-SNAPSHOT:update-property (default-cli) @ parent ---
[INFO] Major version changes allowed
[INFO] Updated ${testng.version} from 5.8 to 5.6
```
or
```
[INFO] --- versions-maven-plugin:2.5-SNAPSHOT:update-property (default-cli) @ module-a2 ---
[INFO] Major version changes allowed
[INFO] Property ${testng.version}: Leaving unchanged as 5.8
```
Than the given property was not found in the reactor. I have to admit it makes more sense to give a warning in such cases (will be a separate issue #171 ).

sregger:
Another area where a warning would be nice is the following

    <properties>
        <my-property.version>1.0.122</my-property.version>
    </properties>

    <dependencyManagement>
        <dependencies>
            
            <dependency>
                <groupId>my.group.id</groupId>
                <artifactId>artifact-A</artifactId>
                <version>${my-property.version}</version>
            </dependency>

            <dependency>
                <groupId>my.group.id</groupId>
                <artifactId>artifact-B</artifactId>
                <version>${my-property.version}</version>
            </dependency>

        </dependencies>
    </dependencyManagement>

> mvn versions:update-property -DnewVersion=1.0.123 -Dproperty=my-property.version

where artifact-A was published for release 1.0.123 but artifact-B was not. In this case the plugin correctly refuses to update the property version. i.e. As this would create an invalid dependency definition for artifact-B.

But it'd be nice for the plugin to print a warning about why the update did not occur. Instead of silently failing. Even if you enable debugging (-X) there is not enough information to tell why the update did not occur. i.e. The messages are the same as when setting an invalid version. This also makes it harder to track down which dependency is no longer published.

Ideally the plugin would print out which dependencies match the new version & which do not match. In general it'd be nice to know why the update did not occur. e.g.

> Property ${my-property.version}: No dependencies match the supplied version leaving unchanged as 1.0.122

> Property ${my-property.version}: Unable to find any repository holding dependencies matching the supplied version leaving unchanged as 1.0.122

> Property ${my-property.version}: Not all dependencies have an artifact matching the supplied version leaving unchanged as 1.0.122
> Matched dependencies: my-group-id:artifact-A, ....
> Unmatched dependencies: my-group-id:artifact-B, ....

khmarbaise:
Can we make a new issue for that please...

sregger:
Submitted https://github.com/mojohaus/versions-maven-plugin/issues/174

Issue #134, opened 2017-03-18
Goal: use-releases doesn't processes imported POMs in dependencyManagement


akacme:
The imported BOMs are not processed correctly by use-releases goal, the fix for #79 is not working either.
I suppose this is due to how maven resolves imported POMs - there is no separate dependency for such import, it's replaced by list of dependencies defined in the POM.

akacme:
Pull Request created: https://github.com/mojohaus/versions-maven-plugin/pull/135

khmarbaise:
Merged PR #135

ajarmoniuk:
I wanted to check if the same issue occurs in other Mojo's since they don't have that fix applied. It turns out that they don't suffer from the issue. 

Further, when I removed the patch from UseReleasesMojo, `import`-scoped dependencies were still present in the collected dependency management dependencies, which means that they are now present out of the box and can be treated by the mojo. And the `import` dependencies get updated. 

Which means that the patch can be removed. I'll create integration tests (important since Maven is involved) before doing that.

ajarmoniuk:
Same issue: #25, #308, #309, #346, #395

ajarmoniuk:
So, this problem is typically resolved by using dependency management from raw model.

PR #135, opened 2017-03-18
Fix for #134: Goal: use-releases doesn't processes imported POMs in dependencyManagement

akacme:
+ added PomHelper.readImportedPOMsFromDependencyManagementSection in order to read imported POMs from raw pom.xml
* UseReleases process imported POMs when processing dependencyManagement
* fixed spelling mistake in javadoc for PomHelperTest


ajarmoniuk:
So far it seems like imported boms are getting processed anyway even if we skip `PomHelper.readImportedPOMsFromDependencyManagementSection`...

PR #145, opened 2017-04-13
Add semantic versioning (semver) goal 

fralalonde:
Updates components of the project's version if it follows the maven-semver format of &lt;major&gt;[.&lt;minor&gt;[.&lt;patch&gt;]][&lt;qualifier&gt;][-SNAPSHOT]

Major, minor and patch versions are zero or positive integers.
Qualifier is any sequence of characters valid in a Maven version.
-SNAPSHOT can be present or absent is not considered part of the qualifier.

Sample valid maven-semver version numbers :
- 0 (major 0, no qualifier, snapshot false)
- 0.2-SNAPSHOT (major 0, minor 2, no qualifier, snapshot true)
- 2.0.6.RELEASE (major 2, minor 0, patch 6, qualifier ".RELEASE", snapshot false)

New version string can be set at once with the `newVersion` parameter.
Individual version numbers can be set to absolute values by one with the `setMajorVersion`, `setMinorVersion`, `setPatchVersion` parameters.
Existing individual version numbers can be incremented by one with the `incrementMajorVersion`, `incrementMinorVersion`, `incrementPatchVersion` parameters.
Snapshot state can be set to an absolute boolean value with the `setSnapshot`parameter.
Qualifier can be set to any absolute string value with the `setQualifier` parameter.

Minor and patch versions numbers can be added to a version if it did not have them before.
Minor and patch versions numbers can not be _removed_ from a version once set.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractModuleVersionUpdaterMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractModuleVersionUpdaterMojo.java
@@ -1,0 +1,323 @@
+package org.codehaus.mojo.versions;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.regex.Pattern;
+
+import javax.xml.stream.XMLStreamException;
+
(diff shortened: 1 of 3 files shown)


fralalonde:
example usage of the `semver` goal :
```
# move back to next snapshot after feature release, reseting patch version
mvn version:semver -DincrementMinorVersion -DsetPatchVersion=0 -DsetSnapshot=true
```

PR #150, opened 2017-05-02
Fixed #153: 'processParent' not working for use-releases goal

### LINKED ISSUE #153: 'processParent' not working for use-releases goal
My expectation with this parameter is that it would look at the parent, and if it was a snapshot version, use the release version.  However, this does not seem to be the case and this setting is just ignored.  The 'processParent' parameter is listed as something that can be used in this goal: http://www.mojohaus.org/versions-maven-plugin/use-releases-mojo.html 

I looked through the code, and couldn't find that this flag was used anywhere in the goal.  Was it a mistake to list that this parameter is used in this goal?

danwelcome:
The documentation listed here makes it seem like this flag should process the parent section of the POM (and remove -SNAPSHOT), but that functionality wasn't there.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDependencyUpdaterMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDependencyUpdaterMojo.java
@@ -237,6 +237,22 @@ public abstract class AbstractVersionsDependencyUpdaterMojo
         return artifact;
     }
 
+    protected String toString( MavenProject project ) {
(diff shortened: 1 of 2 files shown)


khmarbaise:
Would you please so kind to open an issue for that with a descriptive information what the problem is etc. and change the log message to something like this:

```
FIxed #YourIssueNumber
 o Description
```
Thanks for your help.

danwelcome:
I reworded my commit and PR per your request.  Let me know if you need anything else!

PR #165, opened 2017-06-05
Fix #164

### LINKED ISSUE #164: allowMinorUpdates property of update-properties doesn't work correctly.
Version 2.4 introduced allowxxx properties for update-properties.

`mvn versions:update-properties -DallowSnapshots=true -DallowMajorUpdates=false -DallowMinorUpdates=false`

is possible to update a property from 1.0.0 to 1.1.0-SNAPSHOT while minor update is disabled.

Is this behavior intentional?

xjj59307:
- Using snapshot when calculating the upper bound of version range.


xjj59307:
@khmarbaise 

I have tested this commit during our internal release. It was working just fine.
As you mentioned in the issue #164, it's better to use filter to pick the right version. It's a nice-to-have improvement.
Could you think of merging the PR?

rparadowski:
Hi guys, this still isn't merged in 2.7. Change isn't big, but helping a lot. Cheers!

PR #184, opened 2017-07-13
Add Mojo for setting the SCM tag

### LINKED ISSUE #182: Update <scm><tag>...</tag></scm>
I've been using `versions:set` for a long time, and I'm moving away from the `maven-release-plugin` all together. I'm just missing one thing when using `versions:set`. I want to be able to update the `<scm><tag></tag></scm>` entry in my POM.

It would be awesome if I could do this through `versions:set` or another goal in the `versions` plugin.

Any plans to add something like this? Or is this functionality available somewhere else, that I have missed?

anton-johansson:
Closes #182

Usage:
`mvn versions:set-scm-tag -DnewTag=v1.0.0 -DgenerateBackupPoms=false`

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -2,6 +2,10 @@
 
 ## 2.5 (NOT YET RELEASED)
 
+ * [Fixed Issue 182][issue-182]
+
+   Add goal for updating the SCM tag in the POM.
+
  * [Fixed Issue 177][issue-177]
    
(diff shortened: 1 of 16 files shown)


anton-johansson:
Would you also like me to update:
* [ReleaseNotes.md](https://github.com/mojohaus/versions-maven-plugin/blob/master/ReleaseNotes.md)
* GitHub pages

... or is that something you do yourself?

PR #189, opened 2017-07-14
Fix incorrect links in release notes

anton-johansson:


khmarbaise:
@anton-johansson Outch thanks for cleaning this up...

PR #203, opened 2017-09-05
Fix #202 - use-reactor update parents

### LINKED ISSUE #202: -DprocessParent=true is ignored by user-reactor mojo
`processParent` parameter is ignored by user-reactor mojo.

The failing use case is:

1. Hierarchical project structure: A <- B <- C (where A and B are `<packaging>pom</packaging>` and C is `<packaging>jar</packaging>`
1. Parent relationship is on releases (C parent is the latest release of B, and B parent is the latest release of A)
1. Run `mvn versions:use-reactor -DprocessParent=true -DexcludeReactor=false -DallowSnapshots=true` on A (the grandparent).
1. Expected: C parent is updated to point to current version of B in the reactor, same of B.
1. What I get: parents are not updated at all.

amuniz:
See #202 for additional information.

### DIFF
--- a/src/it-repo/module-a1-1.0.pom
+++ b/src/it-repo/module-a1-1.0.pom
@@ -1,0 +1,26 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>localhost</groupId>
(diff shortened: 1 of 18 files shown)


stephenc:
Could you add some integration tests?

stephenc:
See https://github.com/mojohaus/versions-maven-plugin/tree/master/src/it for example integration tests

amuniz:
Ok, will do. Thanks!

amuniz:
@stephenc Tests added.

amuniz:
Build is failing only on JDK7, but the failure seems unrelated to this change:

```
$ mvn --show-version --errors --batch-mode -Prun-its clean verify
Error: JAVA_HOME is not defined correctly.
  We cannot execute /usr/lib/jvm/java-7-oracle/bin/java
```

It is ok on JDK8.

PR #206, opened 2017-09-20
Allow nextSnapshot from a non-snapshot version

aosagie:
Hi,
There's a bug in
```java
mvn versions:set -DnextSnapshot
```
that prevents it from automatically generating the next snapshot if the current version isn't already a SNAPSHOT. This makes it relatively difficult to move to the next snapshot after we've deployed a release.

**Before**
screenshot from 2017-09-20 00-02-59
**After**
screenshot from 2017-09-20 00-01-26

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/SetMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/SetMojo.java
@@ -228,12 +228,10 @@ public class SetMojo
 
         if ( !removeSnapshot && nextSnapshot )
         {
-            boolean havingSnapshot = false;
             String version = getVersion();
             String versionWithoutSnapshot = version;
             if ( version.endsWith( SNAPSHOT ) )
             {
-                havingSnapshot = true;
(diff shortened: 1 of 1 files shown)


bvulaj:
+1 for getting this merged, as we'd also love this feature.

eloucks:
We would also like this feature. Just ran into this issue, came to see if it could be fixed, and there is already a pull request. Please merge!

lion7:
Alternatively maybe @olamy or @hboutemy could take a look at this PR and subsequently PR #255 ?

lion7:
@rfscholte could you take a look at this PR?

olamy:
sounds a good feature. But I'd like to see an it test for this before merging.
Thanks

lion7:
@olamy I created some tests in a separate PR #225. Should I combine this PR and the 1 with the tests in a single PR or would you rather merge them 1 after another?

lion7:
@olamy I created a new PR #284 containing both this fix and the IT tests from PR #225

adamvoss:
#284 merged.  I believe this one can be closed.

Issue #208, opened 2017-09-22
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.6:site (default-site) on project versions-maven-plugin: Err or generating maven-javadoc-plugin:2.10.4:javadoc:


aledell:
Running "mvn site" on a checkout of HEAD at 45d0b9f7901461168d3d2752ffc9bd3928ee7175 fails with 93 javadoc errors.  This is a relatively new complaint since [JDK8 introduced DocLint](https://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete), which is significantly more picky about incomplete javadoc tags like:
* `<code>foo` without a closing `</code>`
* `@param foo` that is not followed by a description
* Methods that throw an exception without having a @throws javadoc

Clearly the correct solution is to fix the javadoc errors.  But I am not terribly familiar with the versions-maven-plugin internals.  I am sure I could provide a patch with obvious fixes, but I would be very nervous about suggesting descriptions to @param elements.

So, rather than fix the 93 things that JDK8's DocLink considers to be javadoc errors, perhaps we should consider disabling the DocLint.  At least temporarily.

According to [this excellent stackoverflow answer](https://stackoverflow.com/a/34591533), we can disable DocLink in a way that is safe for both JDK7 and JDK8 via:
```
  <profiles>
    <profile>
      <!-- Plenty of javadoc errors with JDK8: https://stackoverflow.com/a/34591533 -->
      <id>disable-java8-doclint</id>
      <activation>
        <jdk>[1.8,)</jdk>
      </activation>
      <properties>
        <additionalparam>-Xdoclint:none</additionalparam>
      </properties>
    </profile>
  </profiles>
```

With that <profile/> section added to pom.xml, "mvn site" completes for me, and at least some javadoc for the classes with trouble gets generated into target/site/apidocs/${package}/${class}.html

aledell:
Ugh.  I think I screwed up.  This pull-request fixes #208

aledell:
You might also be interested in this change to mojo-parent: https://github.com/mojohaus/mojo-parent/issues/51

It documents a change in where junit javadoc can be located.

PR #209, opened 2017-09-22
Disable DocLint in JDK7 and JDK8

aledell:
(according to Thiago Porciunclula https://stackoverflow.com/users/2352699/thiago-porci%c3%bancula)


aledell:
Ugh.  I may have screwed up.  This pull-request fixes #208

Issue #220, opened 2017-10-17
Use-Releases with allowRangeMatching should be able to narrow the range to build-numbers only


talksinmath:
We are using the use-releases mojo with the `allowRangeMatching` option turned on.

We do need the option to pad the version used for range matching so that only applicable build numbers are considered even if minor or incremental versions are omitted.

Example:
When using 1.0-SNAPSHOT as a range match, 1.0.1.1 would match. This behaviour is not always intended.
Another expected result would be that only 1.0.0-1 should match (e.g. when 1.0.0 and 1.0.1 are diverging releases).

I will submit a pull request shortly with an implementation that we are using internally already.
Thanks for your consideration.

ajarmoniuk:
I see this has been hanging unresolved for quite some time. Do you still need the feature?

talksinmath:
Nope. Moved on quite some time ago :)
Thanks for asking, forgout about this.

ajarmoniuk:
@slachiewicz can be closed

PR #221, opened 2017-10-17
#220 : use-releases with padVersionForRangeMatching

talksinmath:
This implements the feature proposed in issue #220 

Thanks for your consideration.


talksinmath:
Sorry, somehow I created #220 and #221. That was unintended.

PR #222, opened 2017-10-18
Fixed #194

### LINKED ISSUE #194: User-Defined Version Comparison Methods
The current versions plugin only supports 3 comparison methods (maven, numeric, mercury).  It would be very helpful additional comparison methods could be registered with the VersionComparators and used in version number rules.

Here are a few situations where this feature could be useful: 
* Not all libraries support the standard versioning conventions rendering the plugin useless for those libraries.  Some examples include Hibernate and Spring Framework.  This feature would allow comparators to be defined for them.
* Larger companies may have their versioning conventions for their libraries that don't follow the supported methods.
* ComparableVersion in Maven Artifact is currently not supported by the Versions plugin.

In terms of defining comparison methods, I think it would be reasonable to add a new element in the ruleset xml as follows

```xml
<ruleset comparisonMethod="maven"
         xmlns="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0 http://mojo.codehaus.org/versions-maven-plugin/xsd/rule-2.0.0.xsd">
  <versionComparators>
    <versionComparator name="myVersionComparator" comparatorClass="org.codehaus.mojo.versions.ordering.MyVersionComparator" />
  </versionComparators>
  <rules>
    <rule groupId="*.maven" comparisonMethod="myVersionComparator"/>
  </rules>
</ruleset>
```

This depends on two things currently marked as deprecated: #157 which looks like it's going to stay, and VersionComparators.

I would be willing to try to implement this but I'd like to know if it's even worth pursuing in the first place.

mouyang:
- use Map instead of if-else in VersionComparators.getVersionComparator
- add ruleset.versionComparators to xml
- remove @deprecated from version comparator classes
- update version-rules template with instructions

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/ordering/NumericVersionComparator.java
+++ b/src/main/java/org/codehaus/mojo/versions/ordering/NumericVersionComparator.java
@@ -31,7 +31,6 @@ import java.util.StringTokenizer;
  *
  * @author Stephen Connolly
  * @since 1.0-alpha-3
- * @deprecated
  */
 public class NumericVersionComparator
     extends AbstractVersionComparator
--- a/src/main/mdo/rule.mdo
+++ b/src/main/mdo/rule.mdo
@@ -145,6 +145,16 @@ under the License.
           </association>
         </field>
         <field>
+          <name>versionComparators</name>
+          <version>2.0+</version>
+          <type>List</type>
+          <description>List of VersionComparators to be used in this RuleSet.</description>
+          <association>
+            <type>VersionComparator</type>
+            <multiplicity>*</multiplicity>
+          </association>
+        </field>
+        <field>
           <name>rules</name>
           <required>true</required>
           <version>1.0+</version>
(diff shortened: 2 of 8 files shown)


src/main/java/org/codehaus/mojo/versions/ordering/NumericVersionComparator.java
```diff
@@ -31,7 +31,6 @@
  *
  * @author Stephen Connolly
  * @since 1.0-alpha-3
- * @deprecated
```
khmarbaise: This is not part of the fix...this should be made a different PR...

src/main/mdo/rule.mdo
```diff
@@ -227,5 +237,41 @@ under the License.
         </codeSegment>
       </codeSegments>
     </class>
+    <class xsd.compositor="sequence">
+      <name>VersionComparator</name>
+      <version>2.0+</version>
```
khmarbaise: This mean this breaks compatibility so we need to make change in the version (3.X*) at least...

src/main/java/org/codehaus/mojo/versions/ordering/NumericVersionComparator.java
mouyang: That's fair.  I'll remove it.

src/main/mdo/rule.mdo
mouyang: It does?  Existing rule sets should still work because versionComparator is not marked as required.

mouyang: @khmarbaise I think I answered my own question about compatibility.  I was thinking about schema compatibility when I asked my question, but you were referring to binary compatibility.  Can you confirm/deny/explain?

PR #225, opened 2017-10-23
Added tests that prove removeSnapshot and nextSnapshot

lion7:
The tests in this PR got lost after https://github.com/mojohaus/versions-maven-plugin/pull/105 was closed. 

These tests also proves the issue mentioned in https://github.com/mojohaus/versions-maven-plugin/issues/207 and the fix in https://github.com/mojohaus/versions-maven-plugin/pull/206

### DIFF
--- a/src/it/it-set-010/invoker.properties
+++ b/src/it/it-set-010/invoker.properties
@@ -1,0 +1,3 @@
+invoker.goals=${project.groupId}:${project.artifactId}:${project.version}:set -DremoveSnapshot=true
+invoker.nonRecursive=true
+invoker.buildResult=success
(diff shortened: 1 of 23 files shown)


khmarbaise:
Please make only a pull request which fixes/enhanced a single issue not several in one...

lion7:
The build result breaks because you manually copied the changes of my previous PR #105 in commit eebc0ef without including the accompanying integration tests. This PR actually consists of those same integration tests, and they fail because a new bug was introduced in the previously mentioned commit eebc0ef. 

This newly introduced bug is already reported in https://github.com/mojohaus/versions-maven-plugin/issues/207 and https://github.com/aosagie has provided a fix in PR #206. So for the build result to become green again you would first have to merge PR #206. 

Lastly I was wondering if there is any specific reason for why you didn't merge my previous PR #105 directly but instead copied those changes and committed them with your own name. Credit where credits due?

lion7:
This PR and a proposed fix are combined in PR #284

PR #226, opened 2017-10-25
Cleanup Javadocs

rhwood:
This addresses #208 by:
- adding comments to Javadocs
- not including generated classes in org.codehaus.mojo.versions.model in the Javadocs
- some minor cleanup of existing Javadoc comments

(in a couple of cases, imports were sorted using NetBeans IDE default rules and minor whitespace changes were made)

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -2,6 +2,9 @@
 
 ## 2.5 (NOT YET RELEASED)
 
+ * [Fixed Issue 208][issue-208]
+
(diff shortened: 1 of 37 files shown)


khmarbaise:
Can you please squash your commits so it's a single commit. Furthermore can you make a commit message like:
```
Fixed #208 - Headline
More explanations
```

rhwood:
I tried squashing these into a single commit and it appears that git forced a merge after that. Will open a new PR that is clean.

PR #227, opened 2017-11-02
Issue 195 read enforcer or prerequisites tag

dje1990:
In maven 3.5.0 the prerequisites tag is only valid for plugin type projects.
Currently the versions-maven-plugin requires the prerequisites tag to compare the minimum maven version requirements of plugins in the pom.

If you use the prerequisites tag with maven 3.5.0 you get the following warning:
`[WARNING] The project dje1990:rts:war:HEAD-SNAPSHOT uses prerequisites which is only intended for maven-plugin projects but not for non maven-plugin projects. For such purposes you should use the maven-enforcer-plugin. See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html`

If you remove the prerequisites tag (with any maven version) with the current version of versions-maven-plugin you get:
```
[WARNING] Project does not define minimum Maven version, default is: 2.0
[INFO] Plugins require minimum Maven version of: 3.0
[INFO] 
[ERROR] Project does not define required minimum version of Maven.
[ERROR] Update the pom.xml to contain maven-enforcer-plugin to
[ERROR] force the maven version which is needed to build this project.
[ERROR] See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
[ERROR] Using the minimum version of Maven: 3.0
```

The versions-maven-plugin needs to be able to read the minimum maven version from both the maven-enforcer-plugin and the prerequisites tag (for plugin type projects) for a pom and all it's parent poms.

This pull request attempts to implement the above. The javadoc for the main implementation class reads:

```
/**
 * Finds the minimum Maven version required by a Maven project.
 * Checks for the existence of both the prerequisites.maven property and for maven-enforcer-plugin:enforce goal.
 * Returns null if no minimum version is found.
 * Checks project and it's parents recursively.
 *
 * Pros: works with Maven 3.5.0 which throws a warning if prerequisites.maven is set for a non Maven-Plugin project.
 * Cons: tightly coupled with the maven-enforcer-plugin and the Xpp3Dom configuration tag.
 */
class RequiredMavenVersionFinder {
```

This pull request contains the implementation, unit tests (100% coverage for the main implementation class) and attempts at integration tests (not run yet as I do not have access to the CI). Some manual testing has also been carried out with the following result using maven 3.5.0 with the prerequisites tag removed and the maven-enforcer-plugin needing 3.5.0:
```
[INFO] --- versions-maven-plugin:2.6-SNAPSHOT:display-plugin-updates (version-checks) @ rts ---
[INFO] 
[INFO] 
[INFO] 
[INFO] Project inherits minimum Maven version as: 3.5.0
[INFO] Plugins require minimum Maven version of: 3.0
[INFO] 
[INFO] No plugins require a newer version of Maven than specified by the pom.
```

Please review the change, test the implementation for your own needs, let me know if any additional 
changes are needed, and pull this change onto a feature branch and run the CI.

(21 commits compose of original 7 commits on master, 7 reverts on master, 7 commits on a feature branch).


khmarbaise:
Can you first please squash your commits which makes it possible to make a review on your pull request...Apart from those failing integration tests..

dje1990:
Hi @khmarbaise  I have created a new pull request with a single commit and without the broken integration tests at https://github.com/mojohaus/versions-maven-plugin/pull/230

PR #230, opened 2017-11-08
Fixed 195

dje1990:
Issue 195 read enforcer or prerequisites tag.
https://github.com/mojohaus/versions-maven-plugin/issues/195
https://github.com/mojohaus/versions-maven-plugin/issues/48
https://github.com/mojohaus/versions-maven-plugin/issues/181
https://github.com/mojohaus/versions-maven-plugin/issues/210

In maven 3.5.0 the prerequisites tag is only valid for plugin type projects.
Currently the versions-maven-plugin requires the prerequisites tag to compare the minimum maven version requirements of plugins in the pom.

If you use the prerequisites tag with maven 3.5.0 you get the following warning:
`[WARNING] The project dje1990:rts:war:HEAD-SNAPSHOT uses prerequisites which is only intended for maven-plugin projects but not for non maven-plugin projects. For such purposes you should use the maven-enforcer-plugin. See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html`

If you remove the prerequisites tag (with any maven version) with the current version of versions-maven-plugin you get:
```
[WARNING] Project does not define minimum Maven version, default is: 2.0
[INFO] Plugins require minimum Maven version of: 3.0
[INFO] 
[ERROR] Project does not define required minimum version of Maven.
[ERROR] Update the pom.xml to contain maven-enforcer-plugin to
[ERROR] force the maven version which is needed to build this project.
[ERROR] See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
[ERROR] Using the minimum version of Maven: 3.0
```

The versions-maven-plugin needs to be able to read the minimum maven version from both the maven-enforcer-plugin and the prerequisites tag (for plugin type projects) for a pom and all it's parent poms.

This pull request attempts to implement the above. The javadoc for the main implementation class reads:

```
/**
 * Finds the minimum Maven version required by a Maven project.
 * Checks for the existence of both the prerequisites.maven property and for maven-enforcer-plugin:enforce goal.
 * Returns null if no minimum version is found.
 * Checks project and it's parents recursively.
 *
 * Pros: works with Maven 3.5.0 which throws a warning if prerequisites.maven is set for a non Maven-Plugin project.
 * Cons: tightly coupled with the maven-enforcer-plugin and the Xpp3Dom configuration tag.
 */
class RequiredMavenVersionFinder {
```

This pull request contains the implementation and unit tests (100% coverage for the main implementation class). 
Some manual testing has also been carried out with the following result using maven 3.5.0 with the prerequisites tag removed and the maven-enforcer-plugin needing 3.5.0:
```
[INFO] --- versions-maven-plugin:2.6-SNAPSHOT:display-plugin-updates (version-checks) @ rts ---
[INFO] 
[INFO] 
[INFO] 
[INFO] Project inherits minimum Maven version as: 3.5.0
[INFO] Plugins require minimum Maven version of: 3.0
[INFO] 
[INFO] No plugins require a newer version of Maven than specified by the pom.
```

Please review the change, test the implementation for your own needs, and let me know if any additional 
changes are needed.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
@@ -795,21 +795,8 @@ public class DisplayPluginUpdatesMojo
 
     private String getRequiredMavenVersion( MavenProject mavenProject, String defaultValue )
     {
-        ArtifactVersion requiredMavenVersion = null;
-        while ( mavenProject != null )
-        {
-            final Prerequisites prerequisites = mavenProject.getPrerequisites();
-            final String mavenVersion = prerequisites == null ? null : prerequisites.getMaven();
-            if ( mavenVersion != null )
-            {
-                final ArtifactVersion v = new DefaultArtifactVersion( mavenVersion );
-                if ( requiredMavenVersion == null || requiredMavenVersion.compareTo( v ) < 0 )
-                {
-                    requiredMavenVersion = v;
-                }
-            }
-            mavenProject = mavenProject.getParent();
-        }
+        ArtifactVersion requiredMavenVersion = new RequiredMavenVersionFinder(mavenProject).find();
+
         return requiredMavenVersion == null ? defaultValue : requiredMavenVersion.toString();
     }
 
--- a/src/main/java/org/codehaus/mojo/versions/RequiredMavenVersionFinder.java
+++ b/src/main/java/org/codehaus/mojo/versions/RequiredMavenVersionFinder.java
@@ -95,11 +95,11 @@ class RequiredMavenVersionFinder {
         }
 
         String versionTagValue = versionTag.getValue();
-        if (null == versionTagValue) {
+        if (null == versionTagValue || "".equals(versionTagValue)) {
             return null;
         }
 
-        return new DefaultArtifactVersion(versionTagValue);
+        return processMavenVersionRange(versionTagValue);
     }
 
     private Plugin getMavenEnforcerPlugin(List<Plugin> buildPlugins) {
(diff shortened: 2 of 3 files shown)


glhez:
In the enforcer plugin, you can use version range.
For example, I want to be compatible with the 3.5.x branch of maven and I would use `[3.5.0,3.6.0)`.

Does this take this into account ? (the find method returns only one version but in this case, this should returns _all_ existing versions of maven, therefore it should display for each said version the available plugin updates).

dje1990:
Hi @glhez, version ranges have not been taken into consideration. Further investigation is needed to parse a version range. 
Documentation on version ranges:
https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

Martin-Wegner:
No version support is better than nothing...

digulla:
I would also recommend to support a minimum version first and open another issue to support version ranges.

You may want to have a look at the code in the enforcer plugin to see how it handles these cases.

Alternatively, add a version property which I can specify on the command line. That way, we'd have a quick workaround until the issue can be fixed.

dje1990:
Version ranges are now supported. Please review the commits, test the implementation for your own needs, and let me know if any additional changes are needed.

slachiewicz:
Looks good - verified on maven, spring-boot projects.

dannil:
During the week I encountered a problem with the current code, which is that when you have multiple enforce goals only the first one gets processed, which results in failure for the display-plugin-updates goal. This is being addressed in dje1990/versions-maven-plugin#1, feel free to join in with your own suggestions.

dje1990:
Thanks @dannil , I added minor changes and unit tests in commit 9977213.

khmarbaise:
I would like to merge this within the next few days..is that ok?

dje1990:
Hi @khmarbaise yes ok to merge, there is no known issues with this branch. Thanks

khmarbaise:
Thank for the help and fixing this...

PascalSchumacher:
Thanks you very much for fixing this!

Martin-Wegner:
@khmarbaise Can you tell us about  any release plans, please?

nhojpatrick:
@khmarbaise any update on when this will be released?

digulla:
This was released with version 2.6

PR #239, opened 2017-12-04
Fixed #208 - Cleaned up Javadoc

### LINKED ISSUE #208: [ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.6:site (default-site) on project versions-maven-plugin: Err or generating maven-javadoc-plugin:2.10.4:javadoc:

rhwood:
- Javadocs build on Java 1.8 without lint warnings
- Some Javadoc comments and tags cleaned up
- Javadocs for deprecated generated classes excluded from generation

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -10,6 +10,8 @@
 
  * [Fixed Issue 215][issue-215]
 
+ * [Fixed Issue 208][issue-208]
+
 ## 2.5
 
 
--- a/pom.xml
+++ b/pom.xml
@@ -444,6 +444,15 @@
           <mavenOpts>-Xmx256m</mavenOpts>
         </configuration>
       </plugin>
+      <plugin>
+        <!-- duplicated from reporting to allow configuration to be used with
+          "mvn javadoc:javadoc" as well as "mvn site" -->
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <configuration>
+          <source>${mojo.java.target}</source>
+          <excludePackageNames>org.codehaus.mojo.versions.model</excludePackageNames>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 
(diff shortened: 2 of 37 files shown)


rhwood:
This PR replaces #226 with a single commit.

Issue #237, opened 2017-12-01
Add support for Updating Parent when using UseLatestVersion/Snapshot/Release


DiJu519:
Leverage the isProcessingParent() method in the UseLatestVersion/Snapshot/Release so you can "optionally" treat your parent dependency the same way you treat a normal dependency.

I have a pull request for this. Creating the issue as per community guidelines. Waiting on company approval.

Issue #190, opened 2017-07-18
update-child-modules is doing more than I wish...


chris-nerr:
Hello!
I have a maven project with submodules:

A 1.0.0
- B 1.1.0
- - B-one 1.1.0
- - B-two 1.1.0
- - B-three 1.1.0
- C 1.1.1
- - C-one 1.1.1
- - C-two 1.1.1

All sub-modules of level 3 have the same version than its parent, for example version of module **B-two** is the same than its parent **B**.
If I change the version of my root project **A** to 2.0.0 and I run the command `mvn versions:update-child-modules -DallowSnapshots=true` from the root project then the version of the parent of the 3rd-level modules is updated: version of the C-one parent becomes 2.0.0!!
But that is not what I wish, I just want to update the version of artifact A and its direct children = modules of 2nd level.
Help! :)
Thanks. Chris.

khmarbaise:
Taken a look on http://www.mojohaus.org/versions-maven-plugin/update-parent-mojo.html ? Best would be you make an example project on Github and describe what you've done and then what you like to see...a concrete project is better than describing it in a long abstract...

DiJu519:
I understand the issue, mainly because i'm hitting it as well

Say I have 3 layers of parent inheritance: A, B, and C

A will be an aggregate pom

 A pom.xml

```
  <groupId>com</groupId>
    <artifactId>A</artifactId>
    <version>3.0.0</version>
    <modules>
       <module>B1</module>
        <module>B2</module>
        <module>B3</module>
    </modules>
```

B1 pom.xml
 
```
  <parent>
        <groupId>com</groupId>
    <artifactId>A</artifactId>
    <version>3.0.0-SNAPSHOT</version>
    </parent>

  <groupId>com</groupId>
     <artifactId>B1</artifactId>
     <version>1.0.0-SNAPSHOT</version>

    <modules>
         <module>C1</module>
        <module>C2</module>
         <module>C3</module>
     </modules>
```


 
 C1 pom.xml
 
```
     <parent>
         <groupId>com</groupId>
   <artifactId>B1</artifactId>
    <version>1.0.0-SNAPSHOT</version>
     </parent>
```


Now if i  execute this goal on the A pom.xml, not only will it update the Parent of B1 pom.xml (which is good, because B1's parent is A, and makes sense since the Group and Artifact match

BUT 

[...]

PR #249, opened 2018-01-03
Pull for Issues #190 & #237

DiJu519:
Rebased ontop of master and ran mvn clean verify with IT's enabled. All successful.

Thx.


khmarbaise:
First it's great to produce a pull request. Many thanks for that but it would be better having only a single issue per pull request...

DiJu519:
Is it mandatory to do so..? They are at least separate commits. Pretty straight forward changes.

Also not sure what happened with the TravisCI build, looks to be network related.

DiJu519:
Cancelling this pull request

Need to make a few modifications. Will resubmit.

PR #272, opened 2018-04-26
issue-271 add updateSnapshotsDependencies configuration

alfredyctan:


alfredyctan:
Just found out using use-latest-versions with allowSnapshots = false can archive same behaviour

PR #276, opened 2018-05-22
Fixing issues moving from x.x.x-SNAPSHOT to x.x.x when any usage of MajMiInc filtering is active

DiJu519:
https://github.com/mojohaus/versions-maven-plugin/issues/274

If you have a dependency with the version X.X.X-SNAPSHOT in your POM, and the latest appropriate version you'd expect to move to is X.X.X (sans snapshot), your dependency will not be updated when using the UseLatest series of goals, while specifying a MajMiInc filter.


DiJu519:
@olamy Could you please review this?

PR #289, opened 2018-07-01
Respect the processParent setting by introducing a new base class

### LINKED ISSUE #288: use-next-releases does not respect `processParent` setting
[use-next-releases is documented to have a `processParent` setting](https://www.mojohaus.org/versions-maven-plugin/use-next-releases-mojo.html#processParent).  However, setting the setting to true does not update the parent pom.

A look at the code confirms the Mojo never accesses the `processParent` setting.

adamvoss:
Respect the `processParent` setting by introducing a new base class.

Fixes #288 (`UseNextReleasesMojo`).  Also, fixes an analogous  bug in `ForceReleasesMojo`, `UseDepVersionMojo`, `UseNextReleasesMojo`, `UseNextSnapshotsMojo`, and `UseNextVersionsMojo`

The other existing Mojos that were not modified have their own parent-handling functionality except for `CompareDependenciesMojo` and  `UpdatePropertiesMojo` where the setting still has no effect.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -22,11 +22,19 @@
     is not compatible with project minimum version, not really a proposed
     upgrade
 
- * [Fixed Issue 237][issue-237]
+ * [Fixed Issue 237][issue-237] and [Fixed Issue 288][issue-288]
 
-   Thanks to Julian Di Leonardo <DiJu519@users.noreply.github.com>
+   Thanks to Julian Di Leonardo <DiJu519@users.noreply.github.com> and [Adam Voss](https://github.com/adamvoss)
 
-   Adding parent processing to UseLatestVersion/UseLatestSnapshot/UseLatestRelease
+   Added parent processing to:
+     - UseLatestVersion
+     - UseLatestSnapshot
+     - UseLatestRelease
+     - ForceReleases
+     - UseDepVersion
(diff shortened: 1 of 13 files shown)


DiJu519:
I think i'm responsible for some of the other "parentProcessing" Mojo's. 

One thing i noticed recently, is that if you try to "processParent" but you have no snapshot dependencies, it skips the whole operation. Something i'll have to address, and something to double check works for your change as well :)

nickgr6:
Is there any reason why this fix is not merged yet?

Will this change also fix my issue regarding updating versions (versions:update-properties), in a property when the property is defined in the parent POM, of a project? 

My project structure:
- moduleA
-- pom.xml (uses property from project/pom.xml as dependency <version>
- moduleB
-- pom.xml (uses property from project/pom.xml as dependency <version>
- pom.xml - 

Command which doesn't update the parent pom:
` mvn org.codehaus.mojo:versions-maven-plugin:update-properties -Dincludes=nl.eleven:the-artifact-id:*:3.1.24 -DprocessParent=true -DallowSnapshots=false -DgenerateBackupPoms=false`

There is a higher version (3.1.26) and it does work if a move the property to the poms of moduleA and moduleB (but I don't want to duplicate these versions..)

PR #290, opened 2018-07-02
Move forward to Maven 3 only

### LINKED ISSUE #139: Final Move to Maven 3+ Only
* Change Maven Version 3.0 minimum
* Updated dependencies
* Using maven-artifact-transfer 0.9.0
* Replace old code with generics code.

epabst:
Fixes #139


epabst:
Would you like my commits squashed with your commits?

epabst:
I took the liberty of doing so.  I also reverted some minor changes such as import ordering to make merge conflicts easier for other changes.

pom.xml
```diff
@@ -284,6 +290,11 @@
       <version>1.9.5</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-compat</artifactId>
+      <version>3.0</version>
+    </dependency>
```
khmarbaise: This should not been used cause it contains the compatibility layer to Maven 2...so the plugin should work without this one...

epabst: I found that quite difficult.  Can you help?

dannil: @epabst I assume @khmarbaise means that maven-compat dependency can be dropped; it's a library which supplies Maven 2-specific classes, which I guess this plugin uses, making it able to run under Maven 3 as well. If the plugin is supposed to drop support for Maven 2 that dependency can be safely removed.

epabst:
Without that dependency,  it won't compile.   I don't know how to make it
compile without it although I tried.

Eric

pom.xml
epabst: Without that dependency,  it won't compile.   I don't know how to make it
compile without it although I tried.

dannil: I've tried removing the dependency myself and it seems the biggest offender (among others) is WagonManager. According to https://maven.apache.org/ref/3.5.0/apidocs/org/apache/maven/artifact/manager/WagonManager.html, the whole class has been marked for deprecation; you'll therefore need to find a replacement class which fills the same purpose. A starting point is https://maven.apache.org/ref/3.5.0/apidocs/org/apache/maven/repository/legacy/WagonManager.html, it seems to be the 3.0 equivalent.

dannil: Another alternative could be to refactor the code to use the [Maven Wagon](http://maven.apache.org/wagon/index.html) library with the HTTP provider, but it entails a lot of refactoring. I suggest you wait for input from @khmarbaise.

khmarbaise: If I correctly see the problem you need maven-artifact-transfer which handled the resolving of artifacts See https://maven.apache.org/shared/maven-artifact-transfer/  but I think one functionality is missing in maven-artifact-transfer to get the list of existing versions of a dependency..(Which I'm currently working on..)...

famod: @khmarbaise

> but I think one functionality is missing in maven-artifact-transfer to get the list of existing versions of a dependency..(Which I'm currently working on..)...

Is this done?

PR #314, opened 2018-10-19
Fixed #313 -add runOnlyAtExecutionRoot

### LINKED ISSUE #313: Allow running of set version only at the root of the execution
Using the plugin/goal set-version in a parent pom's profile and running the profile in a child pom propagates the execution to the child pom's children which causes the child pom's children to change version as well.

This ticket is to allow a property to be set that will cause the assembly to run only at the top of a given module tree. That is, run in the project contained in the same folder where the mvn execution was launched.

argenumsky:
This pull request is to allow a property (runOnlyAtExecutionRoot) to be set that will cause the assembly to run only at the top of a given module tree. That is, run in the project contained in the same folder where the mvn execution was launched.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/SetMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/SetMojo.java
@@ -529,14 +529,7 @@ public class SetMojo
         log.debug("Root Folder:" + session.getExecutionRootDirectory());
         log.debug("Current Folder:"+ basedir );
         boolean result = session.getExecutionRootDirectory().equalsIgnoreCase(basedir.toString());
-        if (result)
-        {
-            log.debug( "This is the execution root." );
-        }
-        else
-        {
-            log.debug( "This is NOT the execution root." );
-        }
+        log.debug( "This is the execution root?" + result);
         return result;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/codehaus/mojo/versions/SetMojo.java
```diff
@@ -494,5 +518,26 @@ protected synchronized void update( ModifiedPomXMLEventReader pom )
...
+        log.debug("Current Folder:"+ basedir );
+        boolean result = session.getExecutionRootDirectory().equalsIgnoreCase(basedir.toString());
+        if (result)
+        {
+            log.debug( "This is the execution root." );
+        }
+        else
+        {
+            log.debug( "This is NOT the execution root." );
+        }
```
fabiojb: Another option would be to use ` log.debug("This is the execution root? " + result)` to replace the lines 532 to 539.

argenumsky:
Hi @fabiojb,

Regarding the Integration Test, unfortunately I was not able to follow how to create one based on the source codes. Is there a how-to that I can follow?

fabiojb:
> Hi @fabiojb,
> 
> Regarding the Integration Test, unfortunately I was not able to follow how to create one based on the source codes. Is there a how-to that I can follow?

The new argument `runOnlyAtExecutionRoot` is supposed to be use with the set goal, right?

You can take a look at this `set` goal integration test and create yours based on it:
https://github.com/mojohaus/versions-maven-plugin/tree/master/src/it/it-set-002

PR #317, opened 2018-10-27
More Java 7 related code improvement (20181027)

obfischer:
Serveral minor improvements to the code base as usage of the diamond operator, simplified exception handling and removal of not need try-catch-statements for NullPointerExceptions.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
@@ -177,7 +177,7 @@ public class DisplayPluginUpdatesMojo
                         getProject().getPackaging() } );
                 // we need to provide a copy with the version blanked out so that inferring from super-pom
                 // works as for 2.x as 3.x fills in the version on us!
-                Map<String, String> result = new LinkedHashMap<String, String>( plugins.size() );
+                Map<String, String> result = new LinkedHashMap<>(plugins.size());
                 for ( Plugin plugin : plugins )
                 {
                     result.put( plugin.getKey(), plugin.getVersion() );
@@ -194,7 +194,7 @@ public class DisplayPluginUpdatesMojo
 
                             Pattern pathRegex = Pattern.compile( "/project(/profiles/profile)?"
                                 + "((/build(/pluginManagement)?)|(/reporting))" + "/plugins/plugin" );
-                            Stack<StackState> pathStack = new Stack<StackState>();
+                            Stack<StackState> pathStack = new Stack<>();
                             StackState curState = null;
                             while ( pom.hasNext() )
                             {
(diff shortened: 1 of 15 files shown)


src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -292,7 +292,8 @@ else if ( event.isEndElement() )
     {
         // we want only those parts of pluginManagement that are defined in this project
         Map<String, String> pluginManagement = new HashMap<>();
-        try
+
+        if ( model.getBuild().getPluginManagement() != null )
         {
             for ( Plugin plugin : model.getBuild().getPluginManagement().getPlugins() )
```
fabiojb: You can still get NullPointers here, if `getPlugins()` returns null.

fabiojb: You can still get NullPointers here, if `model.getBuild().getPluginManagement().getPlugins()` returns null.

As long as you removed the try/catch statement, I think there should be a nullity verfiy before the for loop, or put back the try/catch

src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -304,36 +305,21 @@ else if ( event.isEndElement() )
                 }
             }
         }
-        catch ( NullPointerException e )
-        {
-            // guess there are no plugins here
-        }
-        try
+        for ( Profile profile : model.getProfiles() )
```
fabiojb: Same case here, you can still get NullPointers here, if `model.getProfiles()` returns null.

src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -304,36 +305,21 @@ else if ( event.isEndElement() )
...
         {
-            for ( Profile profile : model.getProfiles() )
+            if ( profile.getBuild().getPluginManagement() != null)
             {
-                try
+                for ( Plugin plugin : profile.getBuild().getPluginManagement().getPlugins() )
```
fabiojb: Same case here, you can still get NullPointers here if `profile.getBuild().getPluginManagement().getPlugins()` returns null

src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
obfischer: I checked every getter if it could return `null` or not. A call to `getPlugins()` can never return `null`. Furthermore we would have a lot of `NullPointerExceptions` during the integration tests.

obfischer:
Hi @fabiojb, 

I would like to discuss this pull request again with you, if it is fine for you. When I did my first contribution to the Versions plugin, it was very hard for me to deal with the existing code base. The programming style reminded me of C and was very hard to read. I mean it was very hard to understand the logic. So the goal of this pull request was to make the code easier to read and to understand. Improving the code base would help to get new contributors and to ease the adding of new features.

I am looking forward to your answer and the dicussion with you.

obfischer:
I will close this PR and will start to work on a new one.

fabiojb:
Hi @obfischer sorry for the delayed response. I can still review your code, but I cannot merge it, as I dont have the privileges to.

obfischer:
Hi @fabiojb, no problem at all. I will open an issue and would like to start a discussion how to modernize the code base of the plugin.

bmarwell:
@olamy is just doing this...

https://github.com/mojohaus/versions-maven-plugin/pull/463

I think you will have to rebase *a lot* after his PR is merged. Sorry we didn't aee yours earlier.

PR #365, opened 2019-09-02
Cache result of getBestFitRule

TobiX:
This was identified as a performance hotspot and the result is static
for each run, adding a cache dramatically speeds up some executions.

This is probably not the correct solution, but provides a dramatic speed-up in my cases. While `getBestFitRule` is called only once for each dependency in `dependency-updates-report` and `plugin-updates-report`, it is called over and over in `property-updates-report` (over 38,000,000 times in one of my sample projects). Unfortunately, I got lost in the call-graph pretty quickly - so if someone could prod me into the right direction for a better fix, that would be appreciated.


TobiX:
"Minimal" reproducer project: 
[versions-slow-rules.zip](https://github.com/mojohaus/versions-maven-plugin/files/5169514/versions-slow-rules.zip)

On my system, this takes > 1 minute without my fix and ~ 5 seconds with my fix. If I find some more time I might wrap it into an maven-invoker test (with the timeout feature of MINVOKER-233)

TobiX:
Ah, the attack of the stale bot :/ - FWIW, this is still an issue, we use this patch internally to speed up execution of the report generation...

TobiX:
Added invoker test to demonstrate the problem

PR #369, opened 2019-10-02
Fixed #318

### LINKED ISSUE #318: versions:display-dependency-updates goal need to have excludes and includes property
At present, if the following command is executed:
**mvn versions:display-dependency-updates**
```   <plugin>		
    <inherited>false</inherited>
    <configuration>
    <allowSnapshots>false</allowSnapshots>
    <outputFile>build/outdated-dependencies.txt</outputFile>
    <rulesUri>file:///C:rules.xml</rulesUri>
    <processDependencyManagement>false</processDependencyManagement> 
    <processPluginDependencies>false</processPluginDependencies>  
    </configuration>
    </plugin>
```
**The plugin currently scans for latest updates for all the dependencies listed in the pom file.**
In certain cases, we might need to scan the updates to a limited dependency or a particular classifier or scope.
The display-dependency-updates doesn't currently allow this to be done.
But, other goals like versions:use-latest-versions allows me to have the excludes and includes property in them.

ctasada:
o Added support to filter dependencies while doing `display-dependency-updates`

 Now you can do something like
 ```
 mvn versions:display-dependency-updates -Dincludes=org.springframework.boot
 ```
 Supports both `includes` and `excludes` options. Code is refactored to avoid duplications

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDependencyUpdaterMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDependencyUpdaterMojo.java
@@ -21,7 +21,6 @@ package org.codehaus.mojo.versions;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


slawekjaranowski:
#318 was fixed by #609 

@ctasada - thanks for your effort.

PR #378, opened 2020-01-23
Updated parent version reference

dperezcabrera:
Fixed timeout updating dependencies maven versions through proxy.


dperezcabrera:
It solves this problem: https://stackoverflow.com/questions/57046107/maven-versions-plugin-does-not-respect-proxy-settings

PR #382, opened 2020-02-11
mockito v3.3.3

nhojpatrick:
Updating dependencies, instead of big bang PR, doing each as separate PR


nhojpatrick:
This merge will also merge PR https://github.com/mojohaus/versions-maven-plugin/pull/381 because it reverts the hamcrest v1.x exclusions

PR #385, opened 2020-02-25
junit-jupiter v5.6.2

nhojpatrick:
Updating dependencies, instead of big bang PR, doing each as separate PR

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -271,7 +271,7 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>3.2.4</version>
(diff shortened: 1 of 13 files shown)


nhojpatrick:
This merge will also merge PR https://github.com/mojohaus/versions-maven-plugin/pull/381 because it reverts the hamcrest v1.x exclusions

nhojpatrick:
Requires https://github.com/mojohaus/mojo-parent/pull/81 to be approved, merged and released as I needed to update maven-surefire-plugin and maven-failsafe-plugin to v2.22.2 so mocking would work with junit-jupiter v5.6.0

PR #392, opened 2020-04-20
Fixed #391 - added support for header authorization

### LINKED ISSUE #391: Server configuration httpHeaders ignored
Hello,
I am using version 2.7 to manage dependencies(properties + parent) using the following commands:
- mvn versions:display-parent-updates
- mvn versions:display-property-updates
- mvn versions:update-parent
- mvn versions:update-properties

The project I am using this on has configured a custom repository(Gitlab) which requires a custom header for authorization.

The project pom.xml has a repository(lets say its id is "custom-repo"), than i have a custom server entry in my .m2/settings.xml configured to add a custom HTTP header:
`<server>
      <id>custom-repo</id>
      <configuration>
        <httpHeaders>
          <property>
            <name>Private-Token</name>
            <value>super-secret-token</value>
          </property>
        </httpHeaders>
      </configuration>
    </server>`

I can confirm this setup works when pulling new dependencies, I am encountering only issues when using the versions plugin.

The behavior of the underlying HTTP client was confirmed when i enabled Maven HTTP logging. When Maven pulls new dependencies(not available locally) it adds the custom authorization token to the outgoing requests, when versions plugin checks for new versions - the header isn't added to the request so Gitlab won't accept the request.

Best regards
Etien

eroznik:
Added support for the Maven 3.x repository system, since I didn't want to break the current implementation(backward compatibility for current plugin users) I added a new configuration property to control which version resolution option should be used(useRepositorySystem, default=false).

The changes were tested locally using Maven 3.6.0 on Ubuntu while accessing:
- Maven Central
- self hosted Sonatype Nexus
- self hosted Gitlab (this case is the reason for the pull request/issue)

The observed behavior before the change:
- an artifact package + metadata was deleted from the local repository
- executed `versions:update-properties`, remote requests to the self hosted Gitlab instance failed(404, Gitlab response for unauthorized requests)
- the parent version wasn't changed

The observed behavior after the change, with `useRepositorySystem=true`:
- an artifact package + metadata was deleted from the local repository
- executed `versions:update-properties`, remote requests to the self hosted Gitlab went through, artifact metadata fetched successfully
- the parent version was updated

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,6 +111,7 @@
 
     <mavenVersion>2.2.1</mavenVersion>
     <wagonVersion>2.12</wagonVersion>
+    <aetherVersion>1.0.2.v20150114</aetherVersion>
     <doxiaVersion>1.7</doxiaVersion>
     <doxia-sitetoolsVersion>1.7</doxia-sitetoolsVersion>
     <pluginVersion>${project.version}</pluginVersion>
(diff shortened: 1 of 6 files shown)


eroznik:
@khmarbaise issue summary was updated to `added support for header authorization` which describes exactly what i wanted to accomplish with this pull request. Beside the conflicts that are current present, is there anything else required from my side to proceed with the pull request?

eroznik:
Could a maintainer please look at this? Should the conflicts get resolved so the PR gets approved? Is there something that needs to be changed? Or do we close it and forget about it?

slawekjaranowski:
Looks like fixed in: #825 

@eroznik thanks for your effort

PR #394, opened 2020-05-09
Ensure display output is order, only order at point of printing

nhojpatrick:
I often have to redirect the output into a log file and compare with another project or branch. The output can vary depending as the dependency tree and multi-module project tree changes.

This is the simple version as it only orders the output when printing, instead of doing a wider change and correctly the main collection being used to a SortedSet, as List which currently is used doesn't make sense for me.


khmarbaise:
Thanks for your contribution.

Issue #340, opened 2019-03-15
use-latest-versions ignores dependencies with versions defined as properties


dmsergeevN26:
If you have `<dependency>
                <groupId>a.b.c.d</groupId>
                <artifactId>e</artifactId>
                <version>${e.version}</version>
            </dependency>`
it will be completely ignored by the command

sparsick:
I could observe a different behavior. The goal `use-latest-version` overwrite the version that use a property.

Example:
pom.xml snippet:

```
    <properties>
        <java.version>11</java.version>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <spring-boot.version>2.1.3.RELEASE</spring-boot.version>
    </properties>

<dependencyManagement>
        <dependencies>
            <dependency>
                <!-- Import dependency management from Spring Boot -->
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-dependencies</artifactId>
                <version>2.1.6.RELEASE</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>
    <build>
        <pluginManagement>
            <plugins>
                <plugin>
                    <groupId>org.springframework.boot</groupId>
                    <artifactId>spring-boot-maven-plugin</artifactId>
                    <version>${spring-boot.version}</version>
                </plugin>
           </plugins>
        </pluginManagement>
```

Running `mvn versions:use-latest-versions` results in:

```
[INFO] --- versions-maven-plugin:2.7:use-latest-versions (default-cli) @ demo-app ---
[INFO] Major version changes allowed
[INFO] Updated org.springframework.boot:spring-boot-dependencies:pom:${spring-boot.version} to version 2.1.6.RELEASE
[INFO] Major version changes allowed
```

My expectation is that either the goal ignores versions with properties or it updates the property.

Zegveld:
Ran into the same issue as @sparsick above, created a fix for it, and am already using it locally with much success. See the pull-request for the fix.

PR #396, opened 2020-05-14
Fix for #340.

Zegveld:
#340 issue in short: property versions are overwritten with actual versions instead of leaving them alone.

Also ran into this problem, fixed it by adding an ignore for versions that start with the '${' property reference announcer.

Added it everywhere I also saw the following bit of code that skipped the reactor dependencies:

```
            if ( isExcludeReactor() && isProducedByReactor( dep ) )
            {
                getLog().info( "Ignoring reactor dependency: " + toString( dep ) );
                continue;
            }
```

see also https://github.com/mojohaus/versions-maven-plugin/issues/340


khmarbaise:
Thanks for your contribution.

PR #400, opened 2020-05-20
Fixes #82 ensure version change affects only the current directory (and child directories), not the parent or sibling directories

### LINKED ISSUE #82: [Bug 2.3] Local aggregation root is searched to set the version
I think this peace of code from the SetMojo.java is not correct.
`final MavenProject project =
PomHelper.getLocalRoot( projectBuilder, getProject(), localRepository, null, getLog() );
getLog().info( "Local aggregation root: " + project.getBasedir() );`

I end with my all project modified despite I only wanted to change a one maven module. It automatically searches for the local aggregator root which I think it should not.

stefanseifert:
Fixes #82


elharo:
This change broke at least one of my release scripts. :-(

stefanseifert:
there is a new PR to restore the previous behavior and put the new behavior behind a flag: #427

Issue #442, opened 2021-01-02
NPE upon versionless dependencies in dependencyManagement


ryenus:
When running `mvn versions:resolve-ranges` with versions plugin 2.8.1, it fails with NullPointerException, which seems to be related to dependencies without versions, such as `junit:junit:jar`

<s>
In one of our parent pom, we use org.eclipse.m2e:lifecycle-mapping in a way similar to what described here:
https://stackoverflow.com/a/7423909/537554
</s>

```xml
<dependencyManagement>
   <dependencies>
      <dependency>
         <groupId>junit</groupId>
         <artifactId>junit</artifactId>
      </dependency>
   </dependencies>
</dependencyManagement>
```

```
[ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.8.1:resolve-ranges (default-cli) on project lbn-das-service: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.8.1:resolve-ranges failed.: NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.8.1:resolve-ranges (default-cli) on project lbn-das-service: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.8.1:resolve-ranges failed.
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
[...]

PR #443, opened 2021-01-03
ensure version null safety, fix github #442

ryenus:


ryenus:
@khmarbaise possible to take look? thank you!

PR #446, opened 2021-01-23
[Feature-Request] mvn version:display-dependency-updates --limit <dependency> to check only a specific dependency Fix #212

### LINKED ISSUE #212: [Feature-Request] mvn version:display-dependency-updates --limit <dependency> to check only a specific dependency
It would be great to have an option thats limits version checks for a single dependency (or a group, or a group pattern). 

Especially for large projects or slow repositories it's annoying to check all dependency everytime. 

I will open a PR as soon as possible.

somejavadev:
o Added includes functionality to allow filtering update check by artifact.

The new functionality will decrease the time it takes to check for updates should you wish to only check for specific artifacts.

I have used the same name for the parameter `includes` that you find in the dependency tree plugin to keep consistency.

Example usage:
```bash
mvn versions:display-dependency-updates -Dincludes=org.springframework.boot:spring-boot-starter-data-jpa,org.hibernate
```

Note: 
I might need to get the appropriate code formatter for IntelliJ since I see it seems to format some white space. 

I am not sure if there is an existing utility to do the checks done in the ` private boolean matches( String token, String pattern )` method.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.Arrays;
 
 import javax.xml.stream.XMLStreamException;
 
@@ -55,7 +56,7 @@ import javax.xml.stream.XMLStreamException;
  */
 @Mojo( name = "display-dependency-updates", requiresProject = true, requiresDirectInvocation = false, threadSafe = true )
 public class DisplayDependencyUpdatesMojo
-    extends AbstractVersionsDisplayMojo
+        extends AbstractVersionsDisplayMojo
 {
 
     // ------------------------------ FIELDS ------------------------------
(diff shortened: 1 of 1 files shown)


somejavadev:
Seems like most of the latest PR's are failing with the same 33 errors on the `JDKBuilds / OpenJDK 16 (pull_request) Failing after 8m — OpenJDK 1` build.

src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -148,6 +149,15 @@
     @Parameter( property = "verbose", defaultValue = "false" )
     private boolean verbose;
 
+
+    /**
+     * Limit finding available updates to specific artifacts. Defaults to no limit.
```
olamy, line 154: sounds good but can document what could be the content of this. Maybe with samples. Thanks

src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -487,9 +502,104 @@ private void logUpdates( Map<Dependency, ArtifactVersions> updates, String secti
...
+
+        if ( includes == null )
+        {
+            return dependencies;
+        }
+
+        for ( Dependency dependency : dependencies )
+        {
+
+
```
olamy, line 528: no need of some return here :)

olamy (changes requested):
just some easy changes.
can you please rebase as well to get last changes from master?
Thanks

delanym:
Does this also place the check after the filter for -DincludesList and -Dexcludes/excludesList ?

PR #460, opened 2021-05-11
dereference version properties when working from the original model

jellisgwn:
while looping through dependencies in the original model, dereference any dependency versions that are provided as properties, and replace the version with the property value.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReport.java
+++ b/src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReport.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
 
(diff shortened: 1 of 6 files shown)


slawekjaranowski:
can you provide some it/unit test for confirmation.

jellisgwn:
an it test looks like it shouldn't be an issue. will update PR.

jellisgwn:
@slawekjaranowski added IT for our particular use case.

Our use case: we have a parent POM which defines versions for the the project in `dependencyManagement`, this allows us to define version in the parent and not scatter them through the code. When a dependency is defined for all sub-modules, it gets defined in `dependencies`. In some cases we still use properties to tie together multiple artifacts which share a version in the parent POM. When a dependency is defined in `dependecyManagement` and not in `dependencies` we were not seeing those properties resolved.

This is what we tried to capture in the test, see the usage of `localhost:dummy-impl`.

src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReport.java
```diff
@@ -129,10 +130,24 @@ protected void doGenerateReport( Locale locale, Sink sink )
...
+                // Grab the project properties
+                Properties props = getProject().getProperties();
                 for ( Dependency dep : getProject().getOriginalModel().getDependencyManagement().getDependencies() )
                 {
+                    String version = dep.getVersion();
                     getLog().debug( "Original Dpmg: " + dep.getGroupId() + ":" + dep.getArtifactId() + ":"
-                        + dep.getVersion() + ":" + dep.getType() + ":" + dep.getScope() );
+                        + version + ":" + dep.getType() + ":" + dep.getScope() );
+                    // if the version is referencing a property, dereference it
+                    if ( version.startsWith( "${" ) )
```
bmarwell, line 142: what about versions which are made from more complex strings? e.g. `1.${oldVersion}-SNAPSHOT`?

jellisgwn: Interesting thought. I've not seen that used in the wild, and the current plugin codebase doesn't suggest that it's a supported approach:

https://github.com/mojohaus/versions-maven-plugin/blob/b17a9271478abc12fd33b36d443f476f60005109/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDependencyUpdaterMojo.java#L192

PR #461, opened 2021-06-02
Upgrade parent 63 and Java 1.8

olamy:


bmarwell:
You will need to add a java1.8+ profile for javadoc and the javadoc source: https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#source

For Java 9+ you could also use release: https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#release

pom.xml
```diff
@@ -107,7 +107,7 @@
   </issueManagement>
 
   <properties>
-    <mojo.java.target>1.7</mojo.java.target>
+    <mojo.java.target>1.8</mojo.java.target>
```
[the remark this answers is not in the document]
bmarwell:
```suggestion
    <mojo.java.target>1.8</mojo.java.target>
    <maven.compiler.source>${mojo.java.target}</maven.compiler.source>
```

… at least as long as my mojo-parent-PR is not merged.

.github/workflows/maven.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        java: [8, 11, 14, 15, 16, 17-ea]
+        java: [8, 11]
         jdk: [adopt, zulu]
```
bmarwell, line 29: adopt and zulu are both hotspot, no major differences.
adopt is ambiguous anyway, because you actually mean "adopt-hotspot".
Therefore I would recommend:

```suggestion
        jdk: ['adopt-openj9', zulu]
```

.github/workflows/maven.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        java: [8, 11, 14, 15, 16, 17-ea]
+        java: [8, 11]
```
bmarwell, line 28: I'd leave 16 in and turn that into 17 later in September.
Rationale: Test on all LTS and latest-GA versions. Recommended by Oracle and IBM.
```suggestion
        java: [8, 11, 16]
```

bmarwell (changes requested):

PR #470, opened 2021-07-05
Fixes #469 Support.property file in set-property

### LINKED ISSUE #469: Support `.properties` file input in `set-property`, to update many properties as the same time
I regularly have to update several properties in my project's pom.
Current bad solution is a bash script looping on a .properties file to launch 
`mvn org.codehaus.mojo:versions-maven-plugin:2.8.1:set-property -Dproperty=$propertyName -DnewVersion=$version`
on each property.

I would be much easier and faster to provide a property file in input, like:
myVersions.property
---------------------------
property1=value1
...
propertyN=valueN

and give it in input with `-DpropertyVersionFile=myVersions.property` instead of the `-Dproperty` and `-DnewVersion`

arkel-s:
Fixes #469
Possibility to use set-property with `-DpropertiesVersionsFile` to set multiple properties.

Useful to update values of multiples properties in pom with 1 call.

Example:
  ```properties
  #test.properties
  compo1.version=1.2.3
  compo2.version=4.5.6
  ```

  To update all properties in pom.xml in one call:
  ```
    mvn version:set-property -DpropertiesVersionsFile=test.properties
  ```

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
@@ -19,6 +19,7 @@ package org.codehaus.mojo.versions;
  * under the License.
  */
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Mojo;
@@ -26,8 +27,11 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.codehaus.mojo.versions.api.PomHelper;
 import org.codehaus.mojo.versions.api.PropertyVersions;
 import org.codehaus.mojo.versions.rewriting.ModifiedPomXMLEventReader;
+import org.codehaus.mojo.versions.utils.PropertiesVersionsFileReader;
 
 import javax.xml.stream.XMLStreamException;
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.util.Map;
 
 /**
@@ -35,7 +39,7 @@ import java.util.Map;
  * not build anymore. The sanity checks are done by other goals like <code>update-properties</code> or
  * <code>update-property</code> etc. they are not done here. So use this goal with care.
  *
- * @author Karl Heinz Marbaise
+ * @author Karl Heinz Marbaise          
  * @since 2.5
  */
 @Mojo( name = "set-property", requiresProject = true, requiresDirectInvocation = true, threadSafe = true )
@@ -64,6 +68,15 @@ public class SetPropertyMojo
     private boolean autoLinkItems;
 
     /**
+     * A property file name containing: property=value, to update several properties at the same time.
+     * If 'property' and 'newVersion' are also used, they will be ignored.
+     * @since 2.9
+     */
+
+    @Parameter( property = "propertiesVersionsFile" )
+    private String propertiesVersionsFile;
+
+    /**
      * @param pom the pom to update.
      * @throws MojoExecutionException when things go wrong
      * @throws MojoFailureException when things go wrong in a very bad way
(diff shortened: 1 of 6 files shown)


src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
```diff
@@ -73,24 +85,54 @@
...
+        Property[] propertiesConfig = null;
+        String properties = "";
+        if (!StringUtils.isEmpty(propertiesVersionsFile) ) {
+            logWrongConfigWarning();
+            getLog().debug( "Reading properties and versions to update from file: " + propertiesVersionsFile );
+            PropertiesVersionsFileReader reader = new PropertiesVersionsFileReader(propertiesVersionsFile);
+            try {
+                reader.read();
+            } catch (FileNotFoundException e) {
+                getLog().error("Error while reading property file " + propertiesVersionsFile, e);
```
obfischer, line 97: Just from looking at the code: shouldn't the plugin fail if the properties file is not readable? I guess, to continue without being able to read the file, the result of the operation is unclear.

obfischer (changes requested):

obfischer:
Your pull request is not linked with the issue you are referring to.

src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
```diff
@@ -63,6 +66,15 @@
     @Parameter( property = "autoLinkItems", defaultValue = "true" )
     private boolean autoLinkItems;
 
+    /**
+     * A property file name containing:
+     *     property1=value1
+     *     ...
+     *     propertyN=valueN
+     */
```
obfischer, line 74: Please add the JavaDoc annotation `@Since`, so that users can see since which version the new property is available. I think currently you should set it to 2.9.

src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
```diff
@@ -63,6 +66,15 @@
     @Parameter( property = "autoLinkItems", defaultValue = "true" )
     private boolean autoLinkItems;
 
+    /**
+     * A property file name containing:
```
obfischer, line 70: This is not rendered properly. Please run `mvn site` and check the site generated under `target/site`.

obfischer (changes requested):

obfischer:
BTW, nice addition to the plugin

src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
arkel-s: Oh, right! Fixed by throwing a MojoFailureException

src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
arkel-s: Done.

src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
arkel-s: Fixed!

arkel-s:
Thanks for your very quick feedback @obfischer !

src/test/java/org/codehaus/mojo/versions/utils/PropertiesVersionsFileReaderTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import junit.framework.TestCase;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Test;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.*;
+
+public class PropertiesVersionsFileReaderTest extends TestCase {
```
obfischer, line 11:
```suggestion
public class PropertiesVersionsFileReaderTest {
```
`extends TestCase` is not needed when using JUnit 4

src/test/java/org/codehaus/mojo/versions/utils/PropertiesVersionsFileReaderTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private static final String TEST_PROPERTIES_FILE = "src/test/resources/org/codehaus/mojo/versions/utils/testPropertiesVersionsFile.properties";
+
+    @Test
+    public void testRead() {
+        PropertiesVersionsFileReader reader = new PropertiesVersionsFileReader(TEST_PROPERTIES_FILE);
+        try {
```
obfischer, line 18: I wouldn't catch the execption here. In case of an exception, the test case should fail IMHO

src/test/java/org/codehaus/mojo/versions/utils/PropertiesVersionsFileReaderTest.java
```diff
@@ -0,0 +1,36 @@
+package org.codehaus.mojo.versions.utils;
+
+import junit.framework.TestCase;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Test;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.*;
```
obfischer, line 9:
```suggestion
import org.apache.commons.lang3.StringUtils;
import org.junit.Test;

import java.io.IOException;
import java.util.*;

import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertTrue;
```

obfischer (changes requested):
My last suggestion for changes ;-)

obfischer:
Fixes #469

obfischer:
@arkel-s: Your pull request is not linked with the corresponding issue. Please try to fix this by editing you initial comment and add there `Fixes #469`.

arkel-s:
> @arkel-s: Your pull request is not linked with the corresponding issue. Please try to fix this by editing you initial comment and add there `Fixes #469`.

Done!

ReleaseNotes.md
```diff
@@ -1,5 +1,20 @@
 # Release Notes
 
+* [Fixed Issue 469][issue-469]
```
olamy, line 3: no need of this anymore. Release notes are now automatically generated https://github.com/mojohaus/versions-maven-plugin/releases

src/main/java/org/codehaus/mojo/versions/utils/PropertiesVersionsFileReader.java
```diff
@@ -0,0 +1,59 @@
...
+            propertiesCsv = prop.keySet().stream().map(Object::toString).collect(Collectors.joining(","));
+
+            List<Property> propertiesConfigList = new ArrayList<>();
+            prop.forEach((name, version) ->  {
+                Property propertyConfig = new Property((String)name);
+                propertyConfig.setVersion((String)version);
+                propertiesConfigList.add(propertyConfig);
+            });
+
+            propertiesConfig = propertiesConfigList.toArray(new Property[0]);
```
olamy, line 48: what about something such 
```
            StreamSupport.stream(prop.entrySet().spliterator(), false).map(objectObjectEntry -> {
                Property propertyConfig = new Property((String)objectObjectEntry.getKey());
                propertyConfig.setVersion((String)objectObjectEntry.getValue());
                return propertyConfig;
            }).collect(Collectors.toList()).toArray(new Property[0]);
```

src/test/java/org/codehaus/mojo/versions/utils/PropertiesVersionsFileReaderTest.java
```diff
@@ -0,0 +1,34 @@
+package org.codehaus.mojo.versions.utils;
+
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.*;
```
olamy, line 7: nit we usually avoid star import

olamy (changes requested):
nice work. Thanks to contribute.
Just minor changes maybe?

src/main/java/org/codehaus/mojo/versions/utils/PropertiesVersionsFileReader.java
bmarwell: @olamy you can directly convert into an array, no need to collect it to a list first.
Also, I would like to see the lambda method refactored out for readability. And I guess this code is also used elsewhere.
Rationale: a 1-line map call is much easier to read than a three line lambda.

src/test/java/org/codehaus/mojo/versions/utils/PropertiesVersionsFileReaderTest.java
arkel-s: Done

ReleaseNotes.md
arkel-s: Done, and README updated accordingly

arkel-s:
Thanks for reviewing and merging my PR @olamy and @obfischer :-)
It was my first code contribution to open source.

arkel-s:
Hello @obfischer and @olamy , do you know when a new release of versions-maven-plugin is planned? 
The current one (2.8.1) is more than 1 year old, and I'm looking forward to use my new feature. Thanks :-)

PR #473, opened 2021-07-18
Use JUnit 5 instead of JUnit 4

### LINKED ISSUE #472: Migrate from JUnit 4 to JUnit 5
Writing tests is the hardest part in maintaining long-living software projects. Therefore, we should have the best tools available. So, the versions plugin should use JUnit 5 instead of JUnit 4.

obfischer:
Updated the dependencies to use JUnit 5 instead of JUnit 4. The usage of the JUnit Vintage plattform ensures, that the existing test can be executed without any changes.

Fixed #472

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,7 +98,7 @@
   <properties>
     <mojo.java.target>1.8</mojo.java.target>
     <maven.compiler.source>${mojo.java.target}</maven.compiler.source>
-
+    <junitBomVersion>5.7.2</junitBomVersion>
     <mavenVersion>3.0.5</mavenVersion>
     <wagonVersion>3.4.0</wagonVersion>
     <doxiaVersion>1.9.1</doxiaVersion>
@@ -107,6 +107,18 @@
     <sitePluginVersion>3.7</sitePluginVersion>
     <project.build.outputTimestamp>2020-08-07T21:31:00Z</project.build.outputTimestamp>
   </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.junit</groupId>
+        <artifactId>junit-bom</artifactId>
+        <version>${junitBomVersion}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
 
   <dependencies>
     <dependency>
@@ -227,8 +239,13 @@
       <version>3.8.1</version>
     </dependency>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>


pom.xml
```diff
@@ -99,6 +99,13 @@
     <mojo.java.target>1.8</mojo.java.target>
     <maven.compiler.source>${mojo.java.target}</maven.compiler.source>
 
+    <!--
+     ! Version 64 of the parent POM will provide the dependency
+     ! information for JUnit. Therefore the entry in the dependency management
+     ! section can be removed if version 64 will be released.
+     ! Oliver B. Fischer // 2021-07-18
```
olamy, line 106: no need of this line

olamy:
what is the point if you do migrate code as well?

obfischer:
> what is the point if you do migrate code as well?

I will do it in the one of the next commits. I prefer to do smaller contributions as I can not guarantee when I will be able to continue my work. I think even having the possibility to use JUni 5 will help new contributors. WDYT?

bmarwell:
> > what is the point if you do migrate code as well?
> 
> I will do it in the one of the next commits. I prefer to do smaller contributions as I can not guarantee when I will be able to continue my work. I think even having the possibility to use JUni 5 will help new contributors. WDYT?

Sounds good to me! @olamy I wouldn't be too strict here. Besides, I really think this is a valid and really, really good point.

olamy:
@bmarwell @obfischer I just wonder if we add this dependencies with still old junit dependencies available, contributors may start contributing code with junit5. 
How surefire works when you mix junit4 with junit5?

obfischer:
@olamy It depends on the version of Surefire. The one currently used runs them via JUnit 5. JUnit 5 itself provides the support via a dedicated test engine to run also JUni 4 tests. 

I checked this and Surefire runs the same number of tests after the change as before it.

bmarwell:
Yeah, the Junit5 vintage engine is a drop in replacement. Just make sure we do not downgrade surefire. 😉 I've been using this setup for ages.

olamy:
I will not oppose to this change but well frankly I just feel it weird (even no sense) to make a change to add a dependency which is not used... that's my point.

bmarwell:
Which one is unused? It is replacing junit4. If you remove it, no test will run.

obfischer:
Why it is not used? It runs the tests, allows users to use write tests in a modern way and modernizes the project, as every dependency update it does. It it elementes this modmen "WTF, JUnit 4" 😨

olamy:
As already I mean what is the point of this if you are not replacing `org.junit.*` by `org.junit.jupiter.api.*`
change the code as well otherwise what is the point to only change a dependency...
again feel free to merge I just don't see the point if you do not change any code...

bmarwell:
> I just don't see the point if you do not change any code...

Well, I do. Future tests can be written with junit-jupiter.
Previously you said it is an unused dependency, which is something totally different!

The vintage engine dependency contains `org.junit.*` classes. That is the whole point: No need to rewrite unit tests!

I am going to merge it now.

obfischer:
@olamy I understand your point and as I wrote in one of my comments, I plan to migrate all currently existing test cases. But I decided to split them up in small, but IMHO useful parts, as I can't give any guarantee when I will be able to do it. I don't wan't simply to add to another project a new stale issue, hanging around for some time, which causes only load on the reviewers and maintainers. I hope I was able to illustrate my point. ;-)

bmarwell:
Works for me! :) Quick question: Is there any hurry to convert tests? Maybe we could keep the existing ones for a while unless they need conversions for particular reasons ("don't touch working code")?

obfischer:
I am not in a hurry, but I would like to convert the tests, so that new users will automatically start to write tests using JUnit 5.

bmarwell:
That is not too important: New users would need to create a PR which we could review. ;-)

PR #478, opened 2021-08-10
[Enhancement] Display "Latest" Versions in Reports Summary

### LINKED ISSUE #479: Tedious scrolling to gather latest versions in reports
My usual use case is to update to latest major for security fixes :
updating pom.xml manually from html report.

The tedious part comes from that the information is outside the summary and distributed over a long html page.

Putting latest versions at top in summary could be useful in this usecase and avoid tedious scrolling.

image
image

sultan:
Fixes #479

image
image
image
image

### DIFF
--- a/src/it/it-dependency-updates-report-001/verify.bsh
+++ b/src/it/it-dependency-updates-report-001/verify.bsh
@@ -11,7 +11,7 @@ try
             .replaceAll( "&[^;]+;", " " )
             .replaceAll( "\\s+", " " );
 
-    if ( result.indexOf( "localhost dummy-api 1.1 compile jar 1.1.0-2 1.1.1 1.2 2.0" ) < 0 )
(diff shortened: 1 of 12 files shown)


obfischer (changes requested):
First of all thank you for your contribution. Could you please fix the linking between the pull request and the corresponding issue? This would ease the work of anyone who would like to review your pull request.

sultan:
Fixes #479 
Label enhancment

sultan:
Thanks for the follow up, issue linked in initial comment (#479)

sultan:
The PR fixes the issue by replacing "Next Versions" by "Latest Versions", making the assumption that a user will prefer to update to the Latest Major : eg. updating mysql from 5.1.49 to 8.0.26 (latest).

Updating to older 8.0.7 does not seem like a usefull usecase (hence replacing columns rather than adding)

slachiewicz (approved):

PR #482, opened 2021-08-18
Fixed #481

### LINKED ISSUE #481: Use-dep-version does not process parent version when specified
use-dep-version lists processParent as an accessible parameter but does not actually process the parent version when specified as true.
https://www.mojohaus.org/versions-maven-plugin/use-dep-version-mojo.html#processParent

actsof3000:
Updated use-dep-version to process the parent version when specified

Fixes https://github.com/mojohaus/versions-maven-plugin/issues/481


stefanseifert (changes requested):
thanks for the PR!

can you also provide an integration test to verify the new behavior?
please also merge/rebase with latest develop branch

ajarmoniuk:
I think this PR Is obsolete after #813 
@slachiewicz please close :)

PR #484, opened 2021-09-06
display latest version for artifact as separate mojo

agaengel:


src/main/java/org/codehaus/mojo/versions/DisplayLatestVersionMojo.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+@Mojo(name = "display-latest-version", requiresProject = false, requiresDirectInvocation = true, threadSafe = true)
+public class DisplayLatestVersionMojo
+        extends AbstractVersionsDisplayMojo {
+
+    // ------------------------------ FIELDS ------------------------------
+
+    /**
+     * The artifact to display the latest version found. Must be provided as groupId:artifactId, groupId:artifactId:type or groupId:artifactId:type:classifier
+     * If no scope is provided jar is used.
```
hboutemy, line 52: "if no scope"?
I suppose it is "if no type", isn't it?

src/site/apt/index.apt
```diff
@@ -38,6 +38,8 @@ Versions Maven Plugin
   * {{{./display-dependency-updates-mojo.html}versions:display-dependency-updates}} scans a project's dependencies and
   produces a report of those dependencies which have newer versions available.
 
+  * {{{./display-latest-version-mojo.html}versions:display-latest-version}} displays the latest version for an artefact.
```
hboutemy, line 41: s/artefact/artifact/

PR #527, opened 2022-01-11
Fix detection of plugin updates requiring newer Maven for building than

### LINKED ISSUE #526: Determining the required Maven version for plugins is incorrect
The goal `display-plugin-updates` leverages `RequiredMavenVersionFinder` in 
https://github.com/mojohaus/versions-maven-plugin/blob/04afc2c4b9c98cb9ff7cf4f8e0b7d5678cf58fe7/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java#L825 to determine the minimum required Maven version for a plugin.
That leverages both prerequisites and maven-enforcer-plugin configuration to determine the minimum Maven version and uses the maximum from both in https://github.com/mojohaus/versions-maven-plugin/blob/04afc2c4b9c98cb9ff7cf4f8e0b7d5678cf58fe7/src/main/java/org/codehaus/mojo/versions/RequiredMavenVersionFinder.java#L32.

For plugins only the former does matter, as the latter would only apply if you are interested in building the Maven plugin. For executing maven-plugins only the prerequisites section from the Maven pom.xml counts!

For example the mechanism fails for https://github.com/apache/jackrabbit-filevault-package-maven-plugin/blob/1857619d70082f586c510a71e60934d127151d53/pom.xml#L52 which requires 3.3.9 at run time but 3.6.1 at build time (https://github.com/apache/jackrabbit-filevault/blob/561aaa092d721963f16b860de955e6d1752a350e/parent/pom.xml#L56 and https://github.com/apache/maven-apache-parent/blob/061e19af986a914df4d37b14e7e53a079d9033c5/pom.xml#L337

kwin:
for running

Separate minimal Maven version detection for projects from requirements
of plugins
This closes #526


stefanseifert (commented):
thanks, looks good and solves the problem (i was bothered of it as well).

on remark: you removed the test class RequiredMavenVersionFinderTest completely. probably the test were not all valuable, but especially not covering all the complex parsing code in MinimalMavenBuildVersionFinder.processMavenVersionRange is bothering me - all those variants are not covered in the ITs.

kwin:
> probably the test were not all valuable, but especially not covering all the complex parsing code in MinimalMavenBuildVersionFinder.processMavenVersionRange is bothering me - all those variants are not covered in the ITs.

Most tests were not really useful as were mostly calls on mock objects, but the ones for parsing versions were indeed useful. Let me readd/extend those.

PR #537, opened 2022-01-20
Fixed regression in the maven site rendering

sultan:
https://github.com/mojohaus/versions-maven-plugin/issues/536

Maven Site Plugin and Doxia switching to HTML 5 broke all the reports that includes images.
Including other report plugins.

The proposed fix allows to return back to the original compact format that was more synthetic.


slachiewicz (approved):

michael-o (changes requested):
The comments are redundant because such content is for a commit message.

sultan:
comments removed.

PR #554, opened 2022-02-03
Fix #359

### LINKED ISSUE #359: XML format for property updates report
XML format was introduced for plugin and dependency update reports in #112/#93. 

However it is rather common and convenient to specify version as a property especially when you a working several dependencies of same version (e.g. in pretty much any Spring application).

Goal `versions:property-updates-report` should allow to specify report format(s) similar to `versions:dependency-updates-report` and `versions:plugin-updates-report`.

KTannenberg:
Add ability to generate XML reports in versions:property-updates-report goal

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/PropertyUpdatesReport.java
+++ b/src/main/java/org/codehaus/mojo/versions/PropertyUpdatesReport.java
@@ -19,6 +19,7 @@ package org.codehaus.mojo.versions;
  * under the License.
  */
 
+import java.io.File;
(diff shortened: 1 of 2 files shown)


maxime-michel:
We use this in production since months to create our own dependency bot. It works great!

stefanseifert:
we need an integration test to cover and validation this new behavior - can you provide on in this PR?

here is an example for an existing IT for the old report: https://github.com/mojohaus/versions-maven-plugin/tree/master/src/it/it-property-updates-report-001

Issue #553, opened 2022-02-02
display-dependency-updates should fallback to others scopes if there is no new version


sfragata:
The 3 parameters `allowIncrementalUpdates`, `allowMajorUpdates` and `allowMinorUpdates` are mutually exclusive. 
if I enable major updates (`allowMajorUpdates=true`)and there is no one it should show me the minor or the incremental updates, if `allowMinorUpdates=true` and `allowIncrementalUpdates=true`, The same will be if I enable `allowMinorUpdates` and there is no new version for this scope it should propose incremental one if `allowIncrementalUpdates=true`

ajarmoniuk:
Resolved by #283

ajarmoniuk:
@slachiewicz please close

ajarmoniuk:
The goals will not fall back to other scopes if no version is found -- that would be quite unexpected for most users. However, #283 has clarified the semantics of those switches, which, as you noticed, aren't really independent.

sfragata:
After comment of @ajarmoniuk 
`Resolved by [#283](https://github.com/mojohaus/versions-maven-plugin/issues/283)`
and
`The goals will not fall back to other scopes if no version is found -- that would be quite unexpected for most users. However, [#283](https://github.com/mojohaus/versions-maven-plugin/issues/283) has clarified the semantics of those switches, which, as you noticed, aren't really independent.`

I'm gonna close it

Issue #283, opened 2018-05-30
IMPROVEMENT - allowMinorUpdates=false should imply allowMajorUpdates=false


solter:
I have attached a zip file [mvnVersionsIssues.zip](https://github.com/mojohaus/versions-maven-plugin/files/2054746/mvnVersionsIssues.zip) sufficient for creating a minimal working example of this bug.
It contains two mvn projects, dots1 and dots2, where dots2 depends on dots1.

Here is a complete list of dots1 versions that I have created via modifying the version declaration in the pom and running mvn install without any other changes:
* 3.0.0-0
* 2.1.1-1
* 2.1.0-1
* 2.1.0-0
* 2.0.1-1
* 2.0.1-0
* 2.0.0-11

when dots2 depends on dots1 version 2.0.0-11, and I run the command
    $ mvn -DallowMinorUpdates=false versions:use-latest-versions
the dots1 dependency updates to version 3.0.0-0
Whereias if I run
    $ mvn -DallowMajorUpdates=false -DallowMinorUpdates=false versions:use-latest-versions
then the dots1 dependency correctly updates to version 2.0.1-1

Ideally both commands yield the same update to version 2.0.1-1

DiJu519:
i think this *might* be solved with this 

https://github.com/mojohaus/versions-maven-plugin/issues/274

nsidhaye:
Linking https://github.com/mojohaus/versions-maven-plugin/issues/243

ajarmoniuk:
I think it's a bug and not an enhancement. These allow...updates flags are not orthogonal, but indeed imply themselves. And while it's theoretically possible to e.g. only allow major updates, but not minor updates, it would be highly unintuitive to e.g. only allow updates from 1.1.x to 2.1.x.

PR #555, opened 2022-02-04
display-dependency-updates should fallback to others scopes if there is no new version #553

sfragata:
It will fix the issue
https://github.com/mojohaus/versions-maven-plugin/issues/553


sfragata:
Any news ?

stefanseifert (changes requested):
thanks for the contribution!

a general remark on the integration tests:
* unfortunately the existing code for these allow* properties did not come with any test coverage, so it's very important to add some
* please start the integration test numbers with "010", we have "008" already in the codebase and "009" is on it's way in another PR
* please also add a short suffix after each IT number which gives an indication what's tested in the IT (e.g. the relevant configuration property name)
* if i'm not mistaken the two ITs you added seem to test only the "standard" case - allowMinorUpdates delivers a minor update, testing allowIncrementalUpdates delivers an incremental update. but enhanced the behavior that allowMinorUpdates  should also check for incremental updates if no minor update is available, same applies to allowMajorUpdates.

sfragata:
After comment of @ajarmoniuk 
`Resolved by [#283](https://github.com/mojohaus/versions-maven-plugin/issues/283)`
and
`The goals will not fall back to other scopes if no version is found -- that would be quite unexpected for most users. However, [#283](https://github.com/mojohaus/versions-maven-plugin/issues/283) has clarified the semantics of those switches, which, as you noticed, aren't really independent.`

I'm gonna close it

PR #564, opened 2022-02-17
#563 introduce property "versions.displayTerminalWidth"

### LINKED ISSUE #563: display-dependency/plugin/property-updates: Make terminal width configurable
currently the goals `display-dependency-updates`, `display-plugin-updates` and `display-property-updates` produced formatted output of version lists which fits nicely into a 80 characters terminal window. however, nowadays terminal windows are usually much bigger.

if the output lists dependency with long artifact names the line is split into two lines - and if the output is filtered e.g. with grep the actual version information is cut off in another line.

we should keep the current behavior as default, but make the "target terminal width" size configurable to support widths > 80 chars.

example of the problem when using grep:
```
$ mvn versions:display-dependency-updates | grep conga.definition
[INFO]   io.wcm.devops.conga.definitions:io.wcm.devops.conga.definitions.aem ...
```

stefanseifert:
introduces a new configuration parameter `versions.displayTerminalWidth` for the goals
* `display-dependency-updates`
* `display-plugin-updates`
* `display-property-updates`

to make the display width configurable.

Fixes #563


stefanseifert:
@rfscholte oh, that's even better than making it configurable. will give it a try to make use of this instead of making it configurable in a separate PR.

PR #572, opened 2022-03-09
#571 Make use of MessageUtils.getTerminalWidth when displayTerminalWidth is configured to -1

### LINKED ISSUE #571: display-dependency/plugin/property-updates: Make use of MessageUtils.getTerminalWidth
https://github.com/mojohaus/versions-maven-plugin/pull/564 introduced the feature to make the "displayTerminalWidth" configurable instead of having it fixed to anachronistic 80 chars.

as @rfscholte pointed out there is a method [MessageUtils.getTerminalWidth](https://maven.apache.org/shared/maven-shared-utils/apidocs/org/apache/maven/shared/utils/logging/MessageUtils.html#getTerminalWidth--) which allows to detect the current terminal width if JANSI is available (since Maven 3.5.0).

stefanseifert:
Fixes #571

making use of MessageUtils.getTerminalWidth is _not_ the default behavior. i tested this method on different terminals on windows and got varying results:
* works as expected in powershell and cmd
* does not work in Git bash (the method returned 13249 instead of something about 150 - rendering completely unreadable results).
 
that's probably a bug https://github.com/fusesource/jansi, did not look deeper into this. as Git bash is quite popular on windows i advise not to enable this behavior by default.

@rfscholte WDYT?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -180,6 +180,11 @@
       <version>3.2.0</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.maven.shared</groupId>
+      <artifactId>maven-shared-utils</artifactId>
+      <version>3.3.4</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.maven.wagon</groupId>
       <artifactId>wagon-provider-api</artifactId>
       <version>${wagonVersion}</version>
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDisplayMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsDisplayMojo.java
@@ -26,6 +26,7 @@ import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.shared.utils.logging.MessageUtils;
 import org.codehaus.plexus.util.FileUtils;
 
 /**
@@ -65,6 +66,7 @@ public abstract class AbstractVersionsDisplayMojo
 
     /**
      * Terminal width which should be used to format the padding of the version info list output.
+     * If set to -1 the terminal width is detected automatically (may not be supported with all terminals).
      *
      * @since 2.10.0
      */
@@ -165,6 +167,13 @@ public abstract class AbstractVersionsDisplayMojo
      * @return Offset of the configured display terminal width compared to the default with of 80.
      */
     protected int getDisplayTerminalWidthOffset() {
+        if (this.displayTerminalWidth == -1) {
+            this.displayTerminalWidth = MessageUtils.getTerminalWidth();
+            // getTerminalWidth may return -1 if JANSI is not available
+            if (this.displayTerminalWidth == -1) {
+                this.displayTerminalWidth = DEFAULT_DISPLAY_TERMINAL_WIDTH;
+            }
+        }
         return this.displayTerminalWidth - DEFAULT_DISPLAY_TERMINAL_WIDTH;
     }
 


rfscholte:
Well, Maven will use it as well for its output, so I'd suggest to file an issue at JAnsi, otherwise Maven will hit similar issues

stefanseifert:
that's strange as i'm using the latest version of Maven every day since years with git bash and never have seen such display problems. it seems getTerminalWidth() is not used for the usual use cases.

we also have to consider that the output of these goals may also be redirected to an outputfile, and in that case it might not be desirable to depend on the current terminal width.

stefanseifert:
i've created https://github.com/fusesource/jansi/issues/233 for JANSI

if there are not objections i will merge the PR as-is, having the automatic JANSI terminal width support only optionally if configured explicitly (set to -1), because a) it's broken at least on git bash and windows and b) we have to consider the other output channels as well (e.g. output file).

rfscholte:
Well, it is up to you, I prefer to have a minimum set of parameters where possible. IMO switching to terminalWidth via JAnsi is already an improvement compared to what it was. Even 13249 is likely be better (and should be fixed), as I don't expect output with such long lines.
If you think controlling width of content in a file, don't call it terminalWidth, that's misleading.

stefanseifert:
the problem in this case is that the versions maven plugin uses the width to format the full line filled with "...". so instead of an output like
```
[INFO]   com.github.spotbugs:spotbugs-annotations .............. 4.5.3 -> 4.6.0
```

i get on git bash:
```
[INFO] com.github.spotbugs:spotbugs-annotations ...................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
..................................................................................................................................................... 4.5.3 -> 4.6.0
```

which is completely unusable.

renaming the parameter is a good idea. the output as shown above is not very good readable even if ANSI terminal width is detected correctly because then the lines are rendered too wide in this style.

i will drop this PR and create an new one to rename the parameter to something more feasible.

PR #612, opened 2022-07-04
add spi mechanism for VersionHelper

XenoAmess:
add spi mechanism for VersionHelper

close https://github.com/mojohaus/versions-maven-plugin/issues/606


slawekjaranowski:
Thanks for proposition.

I would do more refactor for such feature
- separate module for api - in order to allow extension to depends only on simple api not on whole plugin implementation
- update plugin dependency to newer Maven api, eg 3.2.5 
- use JSR-330 to provide plugin components

of course step by step

slawekjaranowski:
By the way we should add  IT test for such  feature and documentation with simple example.

XenoAmess:
> Thanks for proposition.
> 
> I would do more refactor for such feature
> 
> * separate module for api - in order to allow extension to depends only on simple api not on whole plugin implementation
> * update plugin dependency to newer Maven api, eg 3.2.5
> * use JSR-330 to provide plugin components
> 
> of course step by step

 That would be a very large refactor...
At least this project would be split to a multi model maven repo.

slawekjaranowski:
> That would be a very large refactor... 
> At least this project would be split to a multi model maven repo.
> Too large that I'm just afraid you would'n review it.

I know that can be a big refactor. 😄

XenoAmess:
> > That would be a very large refactor...
> > At least this project would be split to a multi model maven repo.
> > Too large that I'm just afraid you would'n review it.
> 
> I know that can be a big refactor. 😄

big refactor done! please have a look!

slawekjaranowski:
great - big thanks, 
can you do refactoring in separate PR? - only split by module without other changes.

PR #616, opened 2022-07-12
Fix for issue #582 - update-properties does not work across parent-child pom 

### LINKED ISSUE #582: update-properties does not work across parent-child pom
Version 2.8.1 and above

Parent POM has

```
<properties>
  <version.lib>1.0.0</version.lib>
</properties>

  <dependencies>
    <dependency>
      <artifactId>someLib</artifactId>
      <version>${version.lib}</version>
    </dependency>
  <dependencies>
```
Child POM has

```
<properties>
  <version.lib>1.1.0</version.lib>
</properties>
```
The child $version.lib property is active (i.e. does control the version of someLib in use) however update-properties does not update $version.lib to latest version when run on the child POM. They will do this if the dependency on someLib exists in the child pom.

prodj17:
Fixed #582
 o Updated PomHelper to validate version properties also across parents.

### DIFF
--- a/src/it-repo/dummy-parent-issue-582-1.0.pom
+++ b/src/it-repo/dummy-parent-issue-582-1.0.pom
@@ -1,0 +1,23 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>localhost</groupId>
+  <artifactId>dummy-parent-issue-582</artifactId>
+  <version>1.0</version>
+  <packaging>pom</packaging>
+
+  <properties>
+    <api>1.0</api>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>localhost</groupId>
+      <artifactId>dummy-api</artifactId>
+      <version>${api}</version>
+    </dependency>
+  </dependencies>
+
+
+</project>
--- a/src/it/it-update-properties-issue-582/pom.xml
+++ b/src/it/it-update-properties-issue-582/pom.xml
@@ -1,0 +1,20 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>localhost</groupId>
+        <artifactId>dummy-parent-issue-582</artifactId>
+        <version>1.0</version>
+    </parent>
+
+    <groupId>localhost</groupId>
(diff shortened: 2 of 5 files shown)


prodj17:
@slawekjaranowski do you think you can find some time to look into this?

src/it/it-update-properties-issue-582/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    <version>1.0</version>
+    <packaging>pom</packaging>
+    <name>update-properties with one property, with dependency in parent</name>
+
+    <properties>
+        <api>1.0</api>
+    </properties>
+
+    <build>
+        <pluginManagement>
```
slawekjaranowski, line 21: the same - do we need this?

src/it/it-update-properties-issue-582/verify.bsh
```diff
@@ -0,0 +1,21 @@
+import java.io.*;
```
slawekjaranowski, line 1: please use goovy  script

slawekjaranowski (commented):
- it looks like we can add a unit tests for PomHelper, it should be easier to check coverage by test
- what will be if in parent we will have dependencyManagment and in child dependency without version, and property
- what will be if in parent we will have dependencyManagment and in child no dependency, but only property

disclosure - I'm not know this code ... and logic looks quite complex, so will be good if someone else looks at it

prodj17:
> disclosure - I'm not know this code ... and logic looks quite complex, so will be good if someone else looks at it

The change in logic is actually quite simple, yet for some reason github diff messed up. 

The logic of dependencyManagement vs property stays the same as for a single pom. The only change there is to actually iterate over the parent poms and process those as well when looking for properties.

@olamy @bmarwell I've seen you made changes to PomHelper more recently. Do you guys think you can look into it?

src/it-repo/dummy-parent-issue-582-1.0.pom
```diff
@@ -0,0 +1,105 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>localhost</groupId>
+      <artifactId>dummy-api</artifactId>
+      <version>${api}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <pluginManagement>
```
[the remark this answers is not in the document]
prodj17: Not really. I will remove

src/it/it-update-properties-issue-582/pom.xml
prodj17: same. will remove

src/it/it-update-properties-issue-582/verify.bsh
prodj17: ok

slawekjaranowski:
please change commit message to something more related what we do, eg like:

```
update-properties does not work across parent-child pom
```

pleas also squash to one commit

prodj17:
> please change commit message to something more related what we do, eg like:
> 
> ```
> update-properties does not work across parent-child pom
> ```
> 
> pleas also squash to one commit

@slawekjaranowski done

prodj17:
@slawekjaranowski It has been more than a week and there are no comments. Could we merge this PR?

slawekjaranowski:
Ok.
last request, can you squash to single commit. 
Of course I can do it for you.

slawekjaranowski:
@prodj17 - thanks

PR #618, opened 2022-07-15
Cleanup project pom file

slawekjaranowski:
- use configuration and plugins version from parent
- execute mrm in run-its profile
- remove plexus-container-default from dependencies
- add breadcrumbs for site

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,11 +12,13 @@
   <packaging>maven-plugin</packaging>
 
   <name>Versions Maven Plugin</name>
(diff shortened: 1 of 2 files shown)


bmarwell:
If you merge this, it'll break https://github.com/mojohaus/versions-maven-plugin/pull/619

Do we want to split the plug-in into a multi module project?

slawekjaranowski:
I known that #619 can be break by it.

I think that split project to multi module project will open easy way for creating extension for plugin.

So extension can have only dependency to api not internal plugin implementation and all transitive dependencies.

PR #636, opened 2022-08-16
Adds a `resolve-ranges-historical` goal

mscottford:
This new goal resolves version ranges using historical information. Each resolved version is the latest version that has been published as of the date specified via the value specified in the `versionsAsOf` property. The release history is determined via the [`dependency-history-maven` library](https://github.com/corgibytes/dependency-history-maven/).


slawekjaranowski:
Please describe what case do you try resolve by this change?
Give example of usage.

Issue #539, opened 2022-01-22
support to increase x or y in version of x.y.z, not only z


valuetodays:
Assume current version is 1.0.0-SNAPSHOT.

I can set version to 1.0.1-SNAPSHOTby `mvn versions:set -DnextSnapshot`.
In math, I can set version to the following:
- 1.0.1-SNAPSHOT (finished yet)
- 1.1.0-SNAPSHOT
- 2.0.0-SNAPSHOT

I think it's useful.

ajarmoniuk:
I'll do it

cordahi:
Would it be possible to change the behaviour of this feature to follow standard semantic versioning?
I recently opened https://github.com/mojohaus/versions/issues/1021

If starting from version `1.2.3-SNAPSHOT`,
- `mvn -DnextSnapshot -DnextSnapshotIndexToIncrement=2 versions:set` should result in version `1.3.0-SNAPSHOT`
- `mvn -DnextSnapshot -DnextSnapshotIndexToIncrement=1 versions:set` should result in version `2.0.0-SNAPSHOT`

What is the use case of the current behaviour that preserves the rightmost digits?

PR #642, opened 2022-08-20
support to increase x or y in version of x.y.z, not only z #539

ajarmoniuk:
Implementing the requested feature.

### DIFF
--- a/src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
+++ b/src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
@@ -1,0 +1,76 @@
+package org.codehaus.mojo.versions;
+
+import org.apache.maven.model.Model;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.project.MavenProject;
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.fail;
+
+public class SetMojoTest
+{
+    @Test
+    public void testGetIncrementedVersion() throws MojoExecutionException
+    {
+        new SetMojo()
+        {
+            {
+                assertThat( getIncrementedVersion( "1.0.0", null ), is( "1.0.1-SNAPSHOT" ) );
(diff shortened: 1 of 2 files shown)


slawekjaranowski:
I will move whole block from
```
if ( !removeSnapshot && nextSnapshot )
```

which is responsible to increment snapshot version to separate method - than we can test it by unit tests
It is only operations on Strings - ITs test is to heavy for it

ajarmoniuk:
Suggestion implemented.

src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
```diff
@@ -0,0 +1,76 @@
...
+            {
+                assertThat( getIncrementedVersion( "1.0.0", null ), is( "1.0.1-SNAPSHOT" ) );
+                assertThat( getIncrementedVersion( "1.0.0-SNAPSHOT", null ), is( "1.0.1-SNAPSHOT" ) );
+                assertThat( getIncrementedVersion( "1.0.0-SNAPSHOT", 1 ), is( "2.0.0-SNAPSHOT" ) );
+                assertThat( getIncrementedVersion( "1.0.0-SNAPSHOT", 2 ), is( "1.1.0-SNAPSHOT" ) );
+                assertThat( getIncrementedVersion( "1.0.0-SNAPSHOT", 3 ), is( "1.0.1-SNAPSHOT" ) );
+
+                try
+                {
+                    getIncrementedVersion( "1.0.0", 0 );
```
slawekjaranowski, line 30: should be separate test - another case for check

src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
```diff
@@ -0,0 +1,76 @@
...
+                    fail();
+                }
+                catch ( MojoExecutionException e )
+                {
+                    assertThat( e.getMessage(),
+                            containsString( "nextSnapshotIndexToIncrement cannot be less than 1" ) );
+                }
+                try
+                {
+                    getIncrementedVersion( "1.0.0", 4 );
```
slawekjaranowski, line 40: should be separate test - another case for check

slawekjaranowski:
Next time  please - reformat code in separate PR ...  😄

src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
ajarmoniuk: done

slawekjaranowski (approved):

slawekjaranowski:
Squashed in 540785721e8cb21be5e15dabf031b06eb1ee0125

PR #643, opened 2022-08-20
Require Maven 3.2.5

slawekjaranowski:


ajarmoniuk:
Good idea. That will enable us to use Maven Plugin Testing Harness!

PR #647, opened 2022-08-23
Cache result of getBestFitRule (fixes #575)

### LINKED ISSUE #575: property-updates-report is slow with many dependencies & rules
Not really sure why this happens, but when using a version rules file with the `property-updates-report` target, the report generation is very slow. I tried to implement a workaround in #365 (and added a reproducer today), since I wasn't able to track down the root cause of the issue... If someone has a better idea, I would be open to another approach.

TobiX:
This was identified as a performance hotspot and the result is static
for each run, adding a cache dramatically speeds up some executions.

The included IT shows the problem: Without this optimization, it runs
over a minute, with the fix it finishes instantly.

(Reopened from #365 because I did stupid things with our fork)

### DIFF
--- a/src/it/it-property-updates-report-002-slow/pom.xml
+++ b/src/it/it-property-updates-report-002-slow/pom.xml
@@ -1,0 +1,183 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>localhost</groupId>
+  <artifactId>it-property-updates-report-002-slow</artifactId>
+  <version>1.0</version>
+  <packaging>pom</packaging>
+  <name>property-updates-report</name>
+  <description>Validate performance of property-updates-report</description>
+
+  <properties>
+    <spring.version>3.2.17.RELEASE</spring.version>
+    <hibernate-validator.version>4.3.2.Final</hibernate-validator.version>
+    <slf4j.version>1.7.10</slf4j.version>
+    <jaxb.version>2.2.11</jaxb.version>
+    <jaxb2-basics.version>0.11.1</jaxb2-basics.version>
+    <poi.version>3.17</poi.version>
+    <httpclient.version>4.5.2</httpclient.version>
+    <jackson.version>2.6.7.3</jackson.version>
+    <mariadb-client.version>2.4.3</mariadb-client.version>
+    <postgresql-client.version>42.2.6</postgresql-client.version>
+    <groovy.version>2.0.0</groovy.version>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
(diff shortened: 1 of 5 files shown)


src/it/it-property-updates-report-002-slow/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <artifactId>maven-site-plugin</artifactId>
+          <version>3.7</version>
```
slawekjaranowski, line 189: try @maven-site-plugin.version@  - property defined in parent pom

src/it/it-property-updates-report-002-slow/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+
+  <reporting>
+    <plugins>
+      <plugin>
+        <artifactId>maven-project-info-reports-plugin</artifactId>
+        <version>2.1</version>
+        <reportSets>
+          <reportSet/>
+        </reportSets>
+      </plugin>
```
slawekjaranowski, line 203: we can add: `<excludeDefaults>true</..>` than this will be not needed.

slawekjaranowski (commented):
If we have another IT test which execute: `property-updates-report`

I would consider to remove old, after add moving assertions to this test. 
One test will be enough.

slawekjaranowski:
There is: https://github.com/mojohaus/versions-maven-plugin/tree/master/src/it/it-property-updates-report-001 
can be merged to one test.

Pleas check build status 😄

TobiX:
Hmm, I'm used to write one test per issue, but I can certainly extend the existing test to also show this problem...

slawekjaranowski:
IT tests take long time - I try to minimize duplicate test.
If you think that both tests do something else - ok, we can have its.

TobiX:
I tried to combine both ITs, unfortunately the validation script for the existing test is quite sensitive to dependency changes, so for the sake of stability I'd rather not combine both tests...

slawekjaranowski:
I see a warning in IT build ...

```
[INFO] Building: it-property-updates-report-002-slow/pom.xml
[INFO] artifact org.junit.jupiter:junit-jupiter: checking for updates from central
[INFO] artifact org.junit.jupiter:junit-jupiter-engine: checking for updates from central
Warning:  Could not validate integrity of download from https://repo.maven.apache.org/maven2/org/junit/jupiter/junit-jupiter-engine/maven-metadata.xml.sha1
org.eclipse.aether.transfer.ChecksumFailureException: Checksum validation failed, no checksums available
...
[INFO] artifact org.junit.jupiter.junit-jupiter-engine:junit-jupiter-engine: checking for updates from central
Warning:  /org/junit/jupiter/junit-jupiter-engine/maven-metadata.xml.sha1
java.lang.NullPointerException
    at org.codehaus.mojo.mrm.impl.digest.SHA1DigestFileEntry.<init> (SHA1DigestFileEntry.java:67)
    at org.codehaus.mojo.mrm.impl.digest.SHA1DigestFileEntry$Factory.create (SHA1DigestFileEntry.java:148)
``` 

maybe it is a bug in mrm plugin ... but I would not introduce such warning

can you a litel investigate it?

TobiX:
I just dropped that particular dependency from the test... It's still plenty slow before my change to fail the test before my fix...

slawekjaranowski:
@TobiX - Thanks

Issue #648, opened 2022-08-25
Regression since rev 07042efa: parameter 'localRepository' cannot be null


ajarmoniuk:
Running into the following exception when trying to run depedency-updates-report from command line:

```
Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.11.1-SNAPSHOT:dependency-updates-report failed: The parameter 'localRepository' cannot be null. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.11.1-SNAPSHOT:dependency-updates-report (default-cli) on project versions-maven-plugin: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.11.1-SNAPSHOT:dependency-updates-report failed: The parameter 'localRepository' cannot be null.
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:957)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:289)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:193)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
[...]

PR #650, opened 2022-08-25
Fix for issue #648: de-shadowing the localRepository property

ajarmoniuk:
AbstractMavenReport from maven-reporting-impl:3.2.0 introduced the `localRepository` property, which was supposed to be initialized from a parameter with the same name, and if the same was not supplied, from `${localRepository}`. Unfortunately, it was shadowed by `AbstractVersionsReport`

### DIFF
--- a/src/it/it-abstract-versions-report-001/pom.xml
+++ b/src/it/it-abstract-versions-report-001/pom.xml
@@ -1,0 +1,33 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>localhost</groupId>
+  <artifactId>it-abstract-versions-report-001</artifactId>
+  <version>1.0</version>
+  <packaging>pom</packaging>
+  <name>abstract-versions-report</name>
+  <url>http://localhost/</url>
+
+  <description>
+    Testing command line invocation of dependency- and plugin-updates-report
+  </description>
+
+  <dependencies>
+    <dependency>
+      <groupId>localhost</groupId>
+      <artifactId>dummy-api</artifactId>
+      <version>1.1</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>localhost</groupId>
+        <artifactId>dummy-maven-plugin</artifactId>
+        <version>1.0</version>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/src/it/it-abstract-versions-report-001/verify.groovy
+++ b/src/it/it-abstract-versions-report-001/verify.groovy
@@ -1,0 +1,15 @@
+dependencyUpdatesReport = new File( basedir, "target/site/dependency-updates-report.html" )
+
+assert dependencyUpdatesReport.exists(  )
+
+// some basic (=not comprehensive) checks on the contents of the dependency report
+assert dependencyUpdatesReport.text =~ /\b1\.1\.0-2\b/  // current version of the dependency
+assert dependencyUpdatesReport.text =~ /\b3\.0\b/       // latest major available version
+
+pluginUpdatesReport = new File( basedir, "target/site/plugin-updates-report.html" )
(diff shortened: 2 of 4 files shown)


slawekjaranowski:
Can we add some tests or assertions in existing test to catch such error?

ajarmoniuk:
Ok, I think it doesn't really make sense to create a unit test, and the easiest one would be to use it. If that's not too heavy.

src/it/it-abstract-versions-report-001/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+      <artifactId>dummy-api</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <artifactId>maven-clean-plugin</artifactId>
+          <version>2.2</version>
```
slawekjaranowski, line 42: Please use the latest versions .... in parent pom we have defined properties like: `maven-clean-plugin.version`
so we can use as: 
```
 <version>@maven-clean-plugin.version@ </version>
```

src/it/it-abstract-versions-report-001/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+          <version>2.0</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+    <plugins>
+      <plugin>
+        <groupId>localhost</groupId>
+        <artifactId>dummy-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
```
slawekjaranowski, line 63: What for this?

slawekjaranowski:
Please also add simple assertions to check if report file was created

src/it/it-abstract-versions-report-001/pom.xml
ajarmoniuk: Actually, I wanted to demonstrate that the dependency/plugin update reports work, so we'd actually want to use some older versions, but I think it's better to use mocked artifacts (from the mock repository) instead, like `dummy-maven-plugin`.

src/it/it-abstract-versions-report-001/pom.xml
ajarmoniuk: Plugin from the mock repository, I'll use those instead of some real ones.

ajarmoniuk: Updated the PR. Please re-review. I'm not trying to create a comprehensive test of the dependency/plugin updates reports here, it's just to see if they work at all from command line and produce some sensible output.

slawekjaranowski: ok, it is enough

src/it/it-abstract-versions-report-001/verify.groovy
```diff
@@ -0,0 +1,15 @@
...
+assert dependencyUpdatesReport.text =~ /\b1\.1\.0-2\b/  // current version of the dependency
+assert dependencyUpdatesReport.text =~ /\b3\.0\b/       // latest major available version
+
+pluginUpdatesReport = new File( basedir, "target/site/plugin-updates-report.html" )
+
+assert pluginUpdatesReport.exists(  )
+
+// some basic (=not comprehensive) checks on the contents of the plugin report
+assert pluginUpdatesReport.text =~ /\b2\.5\.3\b/        // current version of the plugin
+assert pluginUpdatesReport.text =~ /\b3\.0\.0-M6\b/     // latest major available version
```
slawekjaranowski, line 15: I would change this assertions to versions of dumy plugin.

Used `3.0.0-M6` is the latest version of m-release-p - when new version will be released - what will be ?

ajarmoniuk: Sorry, my bad. Corrected.

pom.xml
```diff
@@ -361,6 +361,7 @@
           <filterProperties>
             <repository.proxy.url>${repository.proxy.url}</repository.proxy.url>
           </filterProperties>
+          <invokerTest>it-abstract-versions-report-001</invokerTest>
```
slawekjaranowski, line 364: 😄 
you can use property  on cmd `mvn ... -Dinvoker.test=xxx`

slawekjaranowski (commented):
please rebase with the master and squash to one commit

pom.xml
ajarmoniuk: Done.

ajarmoniuk: Done.

Issue #604, opened 2022-06-23
Possible issue which could raise NPE.


sbernard31:
I faced this kind of NPE : 
```
[ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.8.1:set (default-cli) on project leshan: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.8.1:set failed.: NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.8.1:set (default-cli) on project leshan: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.8.1:set failed.
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:957)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:289)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:193)
    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
[...]

PR #655, opened 2022-08-27
Fixing #604: npe in DependencyVersionChanger if artifact version is null

ajarmoniuk:
Fixing an NPE in DependecyVersionChanger if the old version is null + checkstyle fixes + some minor refactoring.

Edit: There's no verification in that test case as the test is considered successful if no exception is raised :-)


ajarmoniuk:
Please also see #656 

This PR can be reworked to incorporate it and use a unit test instead.

src/it/it-set-024-versionless-dependency/child/pom.xml
```diff
@@ -0,0 +1,13 @@
...
+
+  <parent>
+    <groupId>localhost</groupId>
+    <artifactId>test-artifact</artifactId>
+    <version>1.0</version>
+  </parent>
+
+  <artifactId>child</artifactId>
+
+</project>
```
slawekjaranowski, line 13: missing new line 😄

Issue #654, opened 2022-08-26
NPE on versions:set when updating a versionless dependency


ajarmoniuk:
While investigating #604, I came upon a very similar issue, also leading to an NPE, when updating the version of an artifact mentioned in dependencyManagement, without a version in dependencies.

```properties
invoker.goals.1=${project.groupId}:${project.artifactId}:${project.version}:set -DnewVersion=2.0 -DgroupId=localhost -DartifactId=dummy-api
```

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>localhost</groupId>
  <artifactId>it-abstract-versions-report-001</artifactId>
  <version>1.0</version>
  <packaging>pom</packaging>
  <name>abstract-versions-report</name>
  <url>http://localhost/</url>

  <description>
    Testing command line invocation of dependency- and plugin-updates-report
  </description>

  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>localhost</groupId>
        <artifactId>dummy-api</artifactId>
        <version>1.1</version>
      </dependency>
    </dependencies>
  </dependencyManagement>

  <dependencies>
    <dependency>
      <groupId>localhost</groupId>
      <artifactId>dummy-api</artifactId>
    </dependency>
  </dependencies>

</project>
```

And indeed, it leads to a quite similar stack trace (not the same though):

```
[ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.11.1-SNAPSHOT:set (default-cli) on project it-abstract-versions-report-001: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.11.1-SNAPSHOT:set failed. NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.11.1-SNAPSHOT:set (default-cli) on project it-abstract-versions-report-001: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.11.1-SNAPSHOT:set failed.
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
[...]

PR #657, opened 2022-08-27
Fixing #654: NPE in SetMojo when a dependency version is null

ajarmoniuk:
Fixing an NPE when a trying to upgrade a dependency with a version specified in dependencyManagement.

Added some minor refactoring where I modified the files anyway + fixed some obvious vulnerabilities (comparison against a possible null value etc)

### DIFF
--- a/src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
+++ b/src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
@@ -4,14 +4,14 @@ import org.apache.maven.model.Model;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.project.MavenProject;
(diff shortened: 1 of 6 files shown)


src/test/java/org/codehaus/mojo/versions/SetMojoTest.java
```diff
@@ -4,16 +4,14 @@
...
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.fail;
 
-public class SetMojoTest
+public class SetMojoTest extends BaseMojoTestCase
 {
-    @Test
```
slawekjaranowski, line 16: I would not remove `@Test` annotations ... yes I know that TestCase is from junit 3 ...

ajarmoniuk: Ok, restored.

PR #656, opened 2022-08-27
Extension of AbstractMojoTestCase from maven testing harness: injecting projects w/o creating stubs

ajarmoniuk:
Simple extension of `AbstractMojoTestCase`, adding the possibility of injecting a project without having to create project stubs.

Example usage:

```java
    @Test
    public void testVersionlessDependency() throws Exception
    {
        SetMojo myMojo = createMojo( "set", "src/test/resources/org/codehaus/mojo/set/versionless-01/pom.xml" );
        assertNotNull( myMojo );
        myMojo.execute();
    }
```

PR #658, opened 2022-08-28
Fix: making createMojo return Mojo objects (was previously restricted to SetMojo)

ajarmoniuk:
A small correction to #656: it was previously restricted to SetMojo only, now making it more generic. A problem is that we're trying to access a property by name, perhaps there should be a common interface instead, but that's for another PR.

### DIFF
--- a/src/test/java/org/codehaus/mojo/versions/utils/BaseMojoTestCase.java
+++ b/src/test/java/org/codehaus/mojo/versions/utils/BaseMojoTestCase.java
@@ -7,9 +7,9 @@ import java.util.Collections;
 import org.apache.maven.model.Build;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
+import org.apache.maven.plugin.Mojo;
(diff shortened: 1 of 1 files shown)


src/test/java/org/codehaus/mojo/versions/utils/BaseMojoTestCase.java
```diff
@@ -58,11 +58,11 @@ public abstract class BaseMojoTestCase extends AbstractMojoTestCase
      * @return a Mojo instance
      * @throws Exception thrown if mojo lookup fails
      */
-    protected SetMojo createMojo( String goal, String pomFilePath ) throws Exception
+    protected Mojo createMojo( String goal, String pomFilePath ) throws Exception
```
slawekjaranowski, line 61: What do you think:
```
<T extends Mojo> T createMojo(...)
{
 ...
  return (T) mojo;
}
```

ajarmoniuk: Good call. Done.

Issue #213, opened 2017-09-29
versions:use-latest-versions doesn't set previous releases when allowSnapshot=false


chetanddesai:
I have a use-case where I use-latest-versions where I allowSnapshots on my mainline (which works great) but when I set it to try to swap all the snapshots to releases for my release branch using `-DallowSnapshots=false`, it won't "go back in versions" to use the latest release version.

Example:
`pom.xml` has `1.1-20170929132713` but I want to have it update to the latest release candidate which is `1.0`

Is there a way to get this behavior or is it a new feature?

DiJu519:
https://github.com/mojohaus/versions-maven-plugin/issues/78

chetanddesai:
Thanks @DiJu519 - I see`allowDowngrade` in the update-properties documentation, but not yet in use-latest-versions.  http://www.mojohaus.org/versions-maven-plugin/use-latest-versions-mojo.html . Does it work for that as well?

DiJu519:
Last time i was in the code I did not see it implemented.

chetanddesai:
Hopefully it's not too much effort to use the same strategy for some of the other goals.

chetanddesai:
@jdileonardo I think this one is still an issue - I don't see the `allowDowngrade` option, can we reopen it?

ajarmoniuk:
Looking into it.

PR #659, opened 2022-08-28
Implementing #213: new argument: allowDowngrade, which will downgrade a snapshot if these are disallowed

ajarmoniuk:
New argument: `allowDowngrade`, which will downgrade a snapshot version if these are not allowed. Only valid if `allowSnapshots` is `false`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,7 +256,7 @@
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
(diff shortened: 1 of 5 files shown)


ajarmoniuk:
Oh and by the way, I saw quite a few similar copy-pasted classes there, probably some refactoring is due, but that's for another PR.

ajarmoniuk:
Not ready yet.

ajarmoniuk:
To be reopened

PR #660, opened 2022-08-29
Implementing #213: new argument: allowDowngrade, which will downgrade a snapshot if these are disallowed

ajarmoniuk:
New argument: `allowDowngrade`, which will downgrade a snapshot version if these are not allowed. Only valid if `allowSnapshots` is `false`.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
+++ b/src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
@@ -26,6 +26,7 @@ import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.artifact.versioning.VersionRange;
+import org.codehaus.mojo.versions.ordering.InvalidSegmentException;
 import org.codehaus.mojo.versions.ordering.VersionComparator;
 
 /**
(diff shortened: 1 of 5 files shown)


ajarmoniuk:
@slawekjaranowski please review. After this one, we can do the release.

src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
```diff
@@ -64,6 +65,7 @@ protected AbstractVersionDetails()
     {
     }
 
+    @Deprecated
```
slawekjaranowski, line 68: Please give reason, what is replacement for it, and so on

ajarmoniuk: Removed the deprecation instead. Originally added because I'd noticed that the constructor was not being used anymore, but it might come useful for testing.

src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
```diff
@@ -243,9 +220,22 @@ public final ArtifactVersion[] getNewerVersions( String version, boolean include
         return getNewerVersions( new DefaultArtifactVersion( version ), includeSnapshots );
     }
 
+    @Deprecated
```
slawekjaranowski, line 223: one more without comments

ajarmoniuk: Added documentation with explanation for the annotation.

Issue #320, opened 2018-11-28
Some mistake


hjkyoyo:
https://github.com/mojohaus/versions-maven-plugin/blob/1afa5c3c0b5783a7fec30d204832712915317645/src/main/java/org/codehaus/mojo/versions/rewriting/ModifiedPomXMLEventReader.java#L457
It might be something wrong in this code:"int delta = replacement.length() - lastEnd - lastStart;", which result in an exception being thrown when call hasNext():
java.lang.StringIndexOutOfBoundsException: String index out of range: -1971
        at java.lang.AbstractStringBuilder.charAt(AbstractStringBuilder.java:237)
        at java.lang.StringBuilder.charAt(StringBuilder.java:76)
        at org.codehaus.mojo.versions.rewriting.ModifiedPomXMLEventReader.c(ModifiedPomXMLEventReader.java:449)
        at org.codehaus.mojo.versions.rewriting.ModifiedPomXMLEventReader.nextEndIncludesNextEvent(ModifiedPomXMLEventReader.java:438)
        at org.codehaus.mojo.versions.rewriting.ModifiedPomXMLEventReader.hasNext(ModifiedPomXMLEventReader.java:380)
        at org.codehaus.mojo.versions.api.PomHelper.excludeExpressionDependency(PomHelper.java:409)

It might be right as "int delta = replacement.length() - lastEnd + lastStart;". The same mistake is happened in line 558.

ajarmoniuk:
Looking

PR #661, opened 2022-08-29
Fixing #320: Correcting delta computation in ModifiedPomXMLEventReader

ajarmoniuk:
Small fix on delta computation in +replace and +replaceMark methods. Plus a simple unit test.


ajarmoniuk:

Issue #477, opened 2021-08-04
Update the <connection>, <developerConnection> and <url> sections inside <scm>


setedivento:
**[This is a feature request.]**

As per Title, I would like to be able to update the `<connection>`,`<developerConnection>` and `<url>` sections inside the `<scm>` tag in the pom.

It seems you already did something in this direction; in particular, I am referring to this issue: https://github.com/mojohaus/versions-maven-plugin/issues/182,
which corresponding goal is documented here:
https://www.mojohaus.org/versions-maven-plugin/set-scm-tag-mojo.html

What I am asking for is to fill the remaining gap, as indeed pointed out in this comment:
https://github.com/mojohaus/versions-maven-plugin/issues/182#issuecomment-315526534

Maybe a more generic `update-scm` goal could be added?
Or even a series of `set-scm-url`, `set-scm-connection`, `set-scm-developer-connection` goals would be fine.

Thank you very much!

setedivento:
I know SVN is a little bit outdated nowadays, but can someone please have a look at this?
Thank you.

ajarmoniuk:
Sure.

setedivento:
> Sure.

Thanks so much!

PR #664, opened 2022-08-31
New feature (#477): connection, developerConnection, url arguments for set-scm-tag

ajarmoniuk:
Adding the requested new features and modifying the first it for the feature.


slawekjaranowski:
`newTag` parameter is still required, we should require one of allowed parameters

ajarmoniuk:
> `newTag` parameter is still required, we should require one of allowed parameters

Done.

ajarmoniuk:
Please don't merge in yet.

ajarmoniuk:
I have reworked the PR. 

It now lets the user define tags even if they weren't present before or the `scm` element was not present at all.

ajarmoniuk:
Reverted to the old version of the PR: this version retains the original structure of POM. 

Corrected one found issue: newTag would always be applied regardless of being defined or not.

slawekjaranowski (approved):

Issue #231, opened 2017-11-09
Versions:update-parent doesn't have configuration to prevent downgrading


Hi-Fi:
All other update goals have possibility to prevent plugin to downgrade SNAPSHOTs to previous release version (e.g. http://www.mojohaus.org/versions-maven-plugin/update-properties-mojo.html#allowDowngrade). With parent that kind of possibility doesn't exists, and currently it causes parent to be downgraded if child module release (using versions plugin to update SNAPSHOTS) is made before parent's.

```
[INFO] --- versions-maven-plugin:2.5:update-parent (default-cli) @ child-test-module ---
[INFO] Updating parent from 0.2.9-SNAPSHOT to 0.2.8
```

ajarmoniuk:
I'm almost done with this one.

PR #665, opened 2022-08-31
 Fixing #231 bug/feature: added "allowDowngrade", default "false", fixed range handling wrt downgrading

ajarmoniuk:
UpdateParentMojo: added "allowDowngrade", default "false", fixed range handling wrt downgrading. Added a unit test.

This is the same as #662 rebased and squashed (rebasing the original one didn't work for some reason).


ajarmoniuk:

Issue #251, opened 2018-01-10
Invalid segment error when version contains more than 3 segments


jdileonardo:
According to this blog, 
http://blog.soebes.de/blog/2017/02/04/apache-maven-how-version-comparison-works/  

Maven now supports more than 3 segments. However the versions plug-in immediately dies with

`[ERROR] Failed to execute goal use-latest-versions (default-cli) on project mdm: Execution default-cli of goal use-latest-versions failed: Invalid segment, 1, for the 1 segment version: '2.2.38.141' -> [Help 1]`

Any time it comes across a version with more than 3 segments.  I've been mitigating this in the past with a long long list of -Dexcludes , however that's a manual effort.

I'm wondering if @khmarbaise  you have anything planned to address this. Would it be appropriate to instead of fail and throw an error,  WARN that a version has not been updated, or that it has been skipped?

Thx

khmarbaise:
Which version of versions-maven-plugin do you use?

DiJu519:
I've reproduced it with the head of master, so 2.6-SNAPSHOT.  Which i'm just starting to use now.

I am currently on 2.1 but as part of my move to 2.6 i'm also trying to rework my usage of it.

DiJu519:
What i've done locally and temporarily is gone into the 3 VersionComparator classes, commented out the throw to InvalidSegment exception and instead

```
            System.out.println(String.format( "Can't Touch this: Invalid segment, %d, for the %d segment version: '%s'", segment, segmentCount,
                    v.toString() ) );
            return (v);
```
Based on first impressions, i'm liking this behavior more.

larsschaps:
Hi...
i use 2.5 of maven-versions-plugin and have Google Analytics as dependency ( https://mvnrepository.com/artifact/com.google.apis/google-api-services-analytics ).

I get the message:
[ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.5:update-properties (default-cli) on project aldi-bus-bom: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.5:update-properties failed: Invalid segment, 1, for the 1 segment version: 'v3-rev126-1.19.1'

Is this relevant? Is there a work around?

Thanks.

nsidhaye:
Open from Jan. Still nothing... Is this error expected? Are we doing something wrong?

blacelle:
> In my case invalid segment error occurs only if I am ignoring minor & major updates.

[...]

PR #666, opened 2022-09-02
Fixing #251: proper handling of InvalidSegmentException and InvalidVersionSpecificationException

ajarmoniuk:
The situations indicated in the exceptions were not handled properly throughout the code, in some places, `MojoExcecutionException` was simply being thrown leading to the whole run being stopped; `InvalidVersionSpecificationException` was unchecked, leading to the same.

Handling those situations in the Mojo itself rather than in the methods so that a proper handling could be effected where necessary.

Because of all that, a rather sizeable refactoring was due.

More PRs will follow, I want to refactor and polish the code more, but this one is what's necessary for resolving this issue.

So, like I said in the linked issue:

The more I look into it, the more issues I see:

- segment appears to be 0-based (see e.g. `AbstractVersionsUpaterMojo.determineUnchangedSegment` or `MavenVersionComparator.innerIncrementSegment`), but the comparison against the segment count suffers from the 1-off problem:
```java
if ( segment > segmentCount )
{
    throw new InvalidSegmentException( segment, segmentCount,
            version.toString() );
}
```
- segment determination flags (allowMajorUpdates, etc.) are actually not at all independent, but are actually a sort of enum; so an actual enum or an unchanged segment index would be better;
- part of the above: it's not actually possible to have allowMajorUpdates = true and allowMinorUpdates = false; allowMajorUpdates = true implies all others also being true;

There are probably more problems with this, but I'll limit the scope of this item to the issue OP raised only. **The other findings will be done as separate issues.**

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsUpdaterMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsUpdaterMojo.java
@@ -37,6 +37,7 @@ import org.apache.maven.artifact.resolver.ArtifactNotFoundException;
 import org.apache.maven.artifact.resolver.ArtifactResolutionException;
 import org.apache.maven.artifact.resolver.ArtifactResolver;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
+import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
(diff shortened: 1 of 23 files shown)


ajarmoniuk:
@slawekjaranowski please review :smile:

Issue #315, opened 2018-10-20
`set-property` without `newVersion` property doesn't give a useful error message


smithkm:
If the `set-property` goal is executed but `newVersion` isn't set (in my case I had set `new.version` instead) the execution simply fails without any useful error message.  With the `-X` flag the error is shown to be a `NullPointerException` which doesn't give any idea what the problem is,  Giving a useful error that included the name of the missing parameter would make this kind of problem far easier to resolve.

PR #667, opened 2022-09-03
Fixing #315: Guarding against newVersion or property being empty if the properties file is not provided

ajarmoniuk:
Simple fix + unit test

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
@@ -22,7 +22,9 @@ package org.codehaus.mojo.versions;
 import javax.xml.stream.XMLStreamException;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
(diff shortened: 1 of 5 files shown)


Koziol1991:
This is a problematic fix, as now we can't unset property, which is already in pom.xml

slawekjaranowski:
@Koziol-ku thanks for info. 

Please create new issue with example project/pom which can be used to reproduce.

ajarmoniuk:
@Koziol-ku could you please elaborate and, if possible, create a ticket with reproduction steps and pom? The idea was to protect against an NPE which occurred if either `version` or `property` was _empty_.

spyrolabs:
Using `mvn versions:set-property -Dproperty=changelist -DnewVersion=` where we have `<changelist>-SNAPSHOT</changelist>`  would have become `<changelist></changelist>` with version 2.11.0. With version 2.12.0, we have the error explained by @Koziol-ku

ajarmoniuk:
Thanks, I'll fix that.

ajarmoniuk:
Created a PR restoring that.

Issue #258, opened 2018-02-02
Option to ignore versions in pom.xml without need for Rules.xml


simdevmon:
I think it would be convenient if I could specify versions to ignore directly in `pom.xml` under the `configuration` tag. Then I would not have the need to create an extra `Rules.xml` file, but could simply specify everything directly in the pom:

```
<ignoreVersions>
    <ignoreVersion type="regex">.*-beta.</ignoreVersion>
    <ignoreVersion type="regex">.*_ALPHA</ignoreVersion>
</ignoreVersions>
```

sbrunov:
Hi, everyone!

Completely agree with the previous authors.

The requested functionality is quite essential, so it would be great to have it.

Best regards,  
Sergey Brunov.

svbrunov:
Dear maintainers,

Please, consider re-opening this GitHub issue.

Best regards,
Sergey Vyacheslavovich Brunov.

garretwilson:
How can I vote for this? I have a bounty running [on Stack Overflow](https://stackoverflow.com/q/72416739) to try to find a workaround. Most plugins don't make me create an entire separate project just to configure them and have there configurations be present in child projects. Why should this one be any different?

garretwilson:
Could someone give an update on whether there is any thought of fixing this soon? We're about to release a new version of our POM and it would be good to know if this is at least scheduled tentatively or if there isn't any hope for it being addressed.

ajarmoniuk:
It looks like a feature like this has recently been implemented by #369. Please see #318 where it's possible to provide inclusion and exclusion filters for determining which dependency patterns will be considered. Thanks to that, you can rule out patterns such as `.*-beta.` or `.*_ALPHA`, albeit not using regexp, but simple asterisk wildcards.

garretwilson:
@ajarmoniuk , can you clarify? Are these inclusions and exclusions with wildcards something that I can put in the POM? (The examples at #369 only show JVM property usage on the command line.) I don't see this mentioned in the [`display-dependency-updates` documentation](display-dependency-updates).

garretwilson:
Woohoo! @ajarmoniuk you don't know how excited I am to try this out.

> I don't know why plugins and pluginManagement sections have not got filtering

Yeah that's the next thing I thought of to ask you. 😄 I think the dependencies section is the most important, but I'll bet we'll need them in the other sections too.

OK, I'm 🤞 waiting for the release!

[...]

PR #669, opened 2022-09-05
Amended #258: Adding plugin and plugin management dependency includes/excludes

ajarmoniuk:
This simple and short PR extends the closed scope of #258 by adding filtering to plugin and plugin management.

### DIFF
--- a/src/it/it-display-dependency-updates-issue-318-dependencyIncludes-multi/verify.groovy
+++ b/src/it/it-display-dependency-updates-issue-318-dependencyIncludes-multi/verify.groovy
@@ -5,7 +5,6 @@ assert buildLog.text.contains("""
 [INFO]   localhost:dummy-api ....................................... 1.0 -> 3.0
 [INFO]   localhost:dummy-impl ...................................... 1.0 -> 2.2
 [INFO] 
-[INFO] ------------------------------------------------------------------------
 """.replaceAll( "\n", System.lineSeparator() ) )
 
 return true
--- a/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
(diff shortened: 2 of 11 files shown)


src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -257,6 +258,94 @@
     @Parameter( property = "verbose", defaultValue = "false" )
     private boolean verbose;
 
+    /**
+     * Only take these artifacts into consideration.
+     * <p>
+     * Comma-separated list of extended GAV patterns.
+     *
+     * <p>
+     * Extended GAV: groupId:artifactId:version:type:classifier:scope
```
slawekjaranowski, line 267: I'm not sure if scope is used in plugin -> dependencies - IMHO it is always runtime - to check or ask on dev list

src/test/resources/org/codehaus/mojo/display-dependency-updates/issue-258-pom.xml
```diff
@@ -0,0 +1,54 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
slawekjaranowski, line 1: Where this file is used?

src/it/it-display-dependency-updates-issue-318-dependencyIncludes-multi/verify.groovy
```diff
@@ -5,7 +5,6 @@ assert buildLog.text.contains("""
 [INFO]   localhost:dummy-api ....................................... 1.0 -> 3.0
 [INFO]   localhost:dummy-impl ...................................... 1.0 -> 2.2
 [INFO] 
-[INFO] ------------------------------------------------------------------------
```
slawekjaranowski, line 8: Where this log line was generated?

slawekjaranowski: ?

slawekjaranowski: ok, I see more logs is generated in debug level

Issue #632, opened 2022-08-11
Getting valid versions comparison in case of qualifiers (alpha, beta, milestone, rc) 


sultan:
![image](https://user-images.githubusercontent.com/5782559/184096019-377aadb5-06b2-418a-b84d-8d2d39d38c34.png)
![image](https://user-images.githubusercontent.com/5782559/184096088-710930cb-4ebb-4cb3-b64d-bbdac2c93539.png)

In this exemple, we can se how RC versions are treated differently from non RC versions, however Activation APIs are from same 2.0 Major and should be treated as such

sultan:
example of actual problem :

v3.0, searching next minor, creating upper and lower bounds
upper bound included v3.1
lower bound excluded v4.0

however versions v4.0 with qualifiers alpha, beta, rc lies before 4.0, this is an expected normal feature !

the problem is 4.0rc lies between 3.1 and 4.0 excluded
then the problem are the bounds creation that is not well designed

the bounds created should be 3.1snapshot and 4.0snapshot

![image](https://user-images.githubusercontent.com/5782559/188847011-aa41c4dc-2e02-4fae-9aba-b5d6c3152f2b.png)

sultan:
Regression: bug is back after infinity calculus.

sultan:
![image](https://user-images.githubusercontent.com/5782559/195997748-fe02407c-a2ff-46ad-b55c-ba886e8c3638.png)

sultan:
non release versions are now ignored

sultan:
![image](https://user-images.githubusercontent.com/5782559/195997834-0c6b4f13-ef5a-4594-b0b8-33e50c0999c0.png)

sultan:
this leads to contradictory reports : 
![image](https://user-images.githubusercontent.com/5782559/195997874-c02ee379-592e-4b36-8dae-f99d3f52c059.png)

ajarmoniuk:
Created #760

PR #672, opened 2022-09-07
#632 Fixed lower and upper bounds to keep milestones and rcs in the right majors.

sultan:
#632 Fix Bounds so lower and upper bounds keep milestones and rcs in the same majors when they are.

a quick before/after : 

image
image

### DIFF
--- a/src/it/it-dependency-updates-report-001/verify.bsh
+++ b/src/it/it-dependency-updates-report-001/verify.bsh
@@ -17,7 +17,7 @@ try
         System.out.println( "Result = \"" + result +"\"" );
         return false;
     }
-    if ( result.indexOf( "1.1.0-2 Next Version" ) < 0)
+    if ( result.indexOf( "1.1.0-2 Latest Subincremental" ) < 0)
     {
         System.out.println( "Did not identify next version" );
         System.out.println( "Result = \"" + result +"\"" );
--- a/src/it/it-plugin-updates-report-001/verify.bsh
+++ b/src/it/it-plugin-updates-report-001/verify.bsh
@@ -39,7 +39,7 @@ try
         return false;
     }
     if ( result.indexOf( "Group Id localhost Artifact Id dummy-api Current Version 1.1 Classifier Type jar Newer "
-            + "versions 1.1.0-2 Next Version 1.1.1 Next Incremental 1.1.1-2 1.1.2 1.1.3 Latest Incremental 1.2 Next "
+            + "versions 1.1.0-2 Latest Subincremental 1.1.1 Next Incremental 1.1.1-2 1.1.2 1.1.3 Latest Incremental 1.2 Next "
             + "Minor 1.2.1 1.2.2 1.3 Latest Minor 2.0 Next Major 2.1 3.0 Latest Major" ) < 0)
     {
         System.out.println( "Did not identify dependency next versions" );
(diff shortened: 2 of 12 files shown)


sultan:
basically this PR adds a "forceSnapshot" boolean to the increaseSegment method. this allows to create bounds that includes 4.0-snaphost 4.0-rc 4.0-m1 4.0-final in the same 4.0 major, for that we need that boolean to increase 3.0 into lower bound 4.0-snaphost included, and upper bound 5.0-snapshot excluded.

ajarmoniuk:
Hi @sultan do you know if you still have much to do there? I'm working on a refactor which includes changes to UpdateScope (which may lead all the way to getting rid of it). There's a lot of old cruft which needs to be purged. Or maybe we could sync?

Here's the diff:

https://github.com/mojohaus/versions-maven-plugin/compare/master...ajarmoniuk:versions-maven-plugin:aj-5-segments-overhaul?diff=unified

sultan:
All good. No more job needed on my side for this PR. 
I was hoping that these rc ordering things could be implemented in Maven git repo (their artifacts comparator)
If I ever succeed in persuading them to include rc ordering, maybe this job would have to be rethought. 

Agreeing on removing unused method, or a cleanup/refactoring.

sultan:
I can’t have access to my computer this morning. If you can merge both today I’m fine. Though I can help a bit later

ajarmoniuk:
Sure. If you still want to sync, let me know, we can merge our branches.

sultan:
I'll try something in about 6 hours if it can help

ajarmoniuk:
If you go and check in my class, UpdateScope is no more :laughing:

sultan:
great job of factorization. i'll see if i can merge all of it in my PR tomorrow

sultan:
rebased and ready.

sultan:
if you need to merge #708 first, then there is room for me to improve my PR, it seems the forceSnaphsot will be always true after all. then i can simplify everything.

slawekjaranowski:
@sultan #708 merged

sultan:
can Mercury versions allow snapshot qualifier ?
do i need to strip snapshot qualifier before incrementing a NumericVersion ?

sultan:
rebased but requires your attention on above questions before merging

ajarmoniuk:
Looks like it did support -SNAPSHOT and similar qualifiers.

https://www.mail-archive.com/dev@maven.apache.org/msg77826.html

https://www.mail-archive.com/dev@maven.apache.org/msg76344.html

sultan:
the PR currently generates bounds with a "-SNAPSHOT" qualifier, *always*, regardless of comparator (Maven, Mercury, NumericVersion). the current result is tested against all three Test classes which I modified to expect always the snapshot qualifier. if these Tests classes are okay for the reviewers, then the PR can be merged.
my tests on Maven poms are okay.

image

slawekjaranowski:
In the latest picture we have:
```
4.0.0-RC1 Next Major
4.0.0-RC2
```

Should 4.0.0-RC2 will be Next Major instead of 4.0.0-RC1?

sultan:
Not sure about differentiating RC1 vs RC2. the question of differentiating RC1 vs M1 however is another issue that may be adressed in another PR. the problem is caused by dashes and dots not treated the same way. 

The result could better be M1 Next Major followed by RC1 RC2. 

this problem was already there before the PR. I know the involved class but I did not investigate the solution. 

The improvement done in this PR is having M1 RC1 RC2 and Realease labelled in the same major. 

I'm open to improvement if you can provide a textual expected rendering.

sultan:
as a result of fixing #632, it appears we could maybe get rid of the class MajorMinorIncrementalFilter which states that its very existence is filtering milestones and RC from the result of the calls, which are now already filtered out by my PR.

thanks @slawekjaranowski for the merge with the master branch.

i did it on my own branch to avoid losing your push next time.

maybe better for us that i remove MajorMinorIncrementalFilter in another PR ?

Issue #673, opened 2022-09-07
NullpointerException using use-dep-version 2.12.0


franboragina:
Hi all, 

One of our Jenkins build uses:
`mvn versions:use-dep-version -DdepVersion=DEVELOP-SNAPSHOT '-Dincludes=com.jumbotours*:::JET*-SNAPSHOT'`

Since today 07/09/22 (dd/mm/yy)  using 2.12.0  it throws an ugly NullpointerException:

```
[ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.12.0:use-dep-version (default-cli) on project mrsmodel: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.12.0:use-dep-version failed.: NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.12.0:use-dep-version (default-cli) on project mrsmodel: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.12.0:use-dep-version failed.
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:375)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:351)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:171)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:163)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:294)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
[...]

PR #675, opened 2022-09-07
#673: Fixing NPE in UseDepVersionMojo caused by PatternIncludesArtifactFilter expecting a non-null classifier when using wildcards

ajarmoniuk:
Reason for the NPE: the new version of maven-common-artifact-filters, class `PatternIncludesArtifactFilters` (line 593) expects the "artifactoid" to return a non-null value for every field when the pattern contains a wildcard character.

The "artifactoid" is a functional interface basically returning the desired segment (coordinate in the code) of the artifact. The classifer is not set, hence equal null; this leads to NPE.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -186,7 +186,7 @@
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-common-artifact-filters</artifactId>
-      <version>3.3.1</version>
+      <version>3.3.0</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.wagon</groupId>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -274,6 +274,12 @@
       <version>2.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
ajarmoniuk, line 277: Why this: to suppress the SLF4J warning message when using AbstractMojoTestCase (and its children); see https://www.slf4j.org/faq.html

ajarmoniuk:

slawekjaranowski:
For reference https://issues.apache.org/jira/browse/MSHARED-1104

PR #681, opened 2022-09-08
Bumping slf4j

ajarmoniuk:


slawekjaranowski:
It will not works ... Maven support 1.7.x
When we don't use new methods from 2.0 it can be ok ... but generally will brake.

https://issues.apache.org/jira/browse/MNG-7531

slawekjaranowski (changes requested):
.

ajarmoniuk:
Thanks. I think it should be closed until the linked issue is resolved :)

PR #683, opened 2022-09-09
Fixed #682: Restoring the ability to provide an empty "newVersion" argument

### LINKED ISSUE #682: Regression: set-property should allow clearing properties
As indicated in the comments to the PR #667, set-property should allow setting an empty version value to properties.

The NPE coming from ModifiedXMLEventReader is easy to fix. I'll do it tomorrow.

ajarmoniuk:
Looks like folks were using the ability to clear a property value. Restoring it + updating the unit test suite.


ajarmoniuk:

src/test/java/org/codehaus/mojo/versions/SetPropertyMojoTest.java
```diff
@@ -39,26 +42,30 @@ public void testNullNewVersion()
...
-            fail();
-        }
-        catch ( MojoExecutionException e )
-        {
-            assertThat( e.getMessage(),
-                    containsString( "Invalid execution arguments: newVersion must not be empty" ) );
-        }
+
+        setVariableValueToObject( mojo, "property", "dummy-api-version" );
+        setVariableValueToObject( mojo, "newVersion", null );
```
slawekjaranowski, line 49: What with case when we have empty string? like for `-DnewVersion=`

ajarmoniuk: Should be the same, but added a test case just to be safe. Please refresh.

ajarmoniuk:
Could you restart the jobs?

Issue #684, opened 2022-09-09
Option to ignore versions in pom.xml without need for Rules.xml


garretwilson:
#258 requested the ability to allow `<ignoreVersions>` or its equivalent to be specified in the POM, like most other plugins, rather than in some separate, external rules file which is not inherited by the POM. #258 implemented a new exclusion system in the POM, but the implementation was incomplete—only some of the checked versions are addressed. Specifically no exclude functionality was made for the `display-plugin-updates` goal

This ticket is to complete adding exclusion functionality equivalent to `<ignoreVersions>` in the POM.

ajarmoniuk:
Sorry, I believe I misunderstood you in the beginning. The other ticket implemented a presentation filter of the listed dependency updates.

What I'm going to do now is create an alternative for providing a ruleset in the rules.xml file, via properties. This will be available (almost) universally.

slawekjaranowski:
I'm starting think, that we implements the same things in duple place ...

I would do an exercise - add parameters with type `RuleSet` together with `rulesUri`. 
We can use feature of mapping complex object of Maven:
https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Mapping_Complex_Objects

When `DefaultVersionsHelper` will receive filled object simply can use it instead of load one from uri.

ajarmoniuk:
Excellent idea.

garretwilson:
> Sorry, I believe I misunderstood you in the beginning. The other ticket implemented a presentation filter of the listed dependency updates.

First of all, sorry for the initial grumpy response earlier. I woke up expecting this to finally work and was hit with bad news in my inbox first thing in the morning. 😬 

I think there will be less misunderstanding if we first clarify the use case before compiling a long list of features.

_Use case: we want all of our applications, simply by inheriting from a common parent POM, with no further configuration, to be able to check, using a single command, whether any of their dependencies or plugins have a later release version (i.e. not release candidate, not milestone, etc.) available._

Also see my [Versions Maven Plugin rules that are inheritable](https://stackoverflow.com/questions/72416739/versions-maven-plugin-rules-that-are-inheritable) question on Stack Overflow.

> What I'm going to do now is create an alternative for providing a ruleset in the rules.xml file, via -D properties or configuration entries in the pom.

That's what I thought this #258 was for! 😅 

[...]

PR #686, opened 2022-09-12
Implementing #684: allow providing ruleSet in POM

ajarmoniuk:
Providing the possibility to provide the ruleSet via POM. As an extra which comes at low cost, it will also be possible to provide just the global ignored versions via the `maven.version.ignore` property (property name chosen to be in line with similar properties for this plugin).

Added some unit tests where it didn't come at a great runtime cost, in other cases created integration tests to test as many mojos as possible where the change occurs.

On top of that, some very minor refactoring and cleanup.


src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -292,6 +302,39 @@ private static RuleSet loadRuleSet( String serverId, Settings settings, WagonMan
         return ruleSet;
     }
 
+    @SuppressWarnings( "checkstyle:AvoidNestedBlocks" )
+    private static RuleSet loadRuleSet( Collection<String> ignoredVersions, RuleSet originalRuleSet )
+    {
+        RuleSet ruleSet = new RuleSet();
+        if ( originalRuleSet != null )
+        {
```
slawekjaranowski, line 310: What with ` originalRuleSet.getComparisonMethod()` ?

src/site/apt/version-rules.apt.vm
```diff
@@ -82,11 +88,82 @@ Version number rules
...
+  <build>
+    ...
+    <plugins>
+      ...
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>versions-maven-plugin</artifactId>
+        <version>${pluginVersion}</version>
+        <configuration>
+          <ruleSet>
```
slawekjaranowski, line 110:
```
            <comparisonMethod></..>
```

src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -215,6 +218,13 @@ public DefaultVersionsHelper( RepositorySystem repositorySystem, ArtifactResolve
         this.log = log;
     }
 
+    /**
+     * Private constructor used by the builder
+     */
+    private DefaultVersionsHelper()
```
slawekjaranowski, line 224: One constructor and everything should be done in builder

slawekjaranowski (changes requested):
We have two methods `loadRuleSet` in `DefaultVersionsHelper`
one is called from constructor one from builder 
there is logic for checking `ignoredVersions` and 
there is logic for checking `ignore Versions` and creating / coping ruleSet depends on it ... it looks a little complicated

Maybe simply in order:
 - if ruleSet exist in plugin configuration use it
 - if rulesUri exist use it
 - always add ignoredVersions to ruleSet

src/it/it-compare-dependencies-001/invoker.properties
```diff
@@ -1,2 +1,5 @@
-invoker.goals=${project.groupId}:${project.artifactId}:${project.version}:compare-dependencies
-invoker.systemPropertiesFile = test.properties
+invoker.goals.1 = ${project.groupId}:${project.artifactId}:${project.version}:compare-dependencies
+invoker.mavenOpts.1 = -DreportOutputFile=depDiffs-1.txt -DremotePom=localhost:dummy-bom-pom:1.0
```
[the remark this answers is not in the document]
ajarmoniuk: It could, but I wanted to be consistent -- most tests split these so that it's easier to spot the options :)

src/it/it-plugin-updates-report-issue-684-001/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+            <reports>
+              <report>plugin-updates-report</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+      </plugin>
+    </plugins>
+  </reporting>
+
+</project>
```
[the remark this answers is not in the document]
ajarmoniuk: Didn't know about that rule -- added newlines where they were missing.

src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
ajarmoniuk: Done.

src/site/apt/version-rules.apt.vm
ajarmoniuk: it's there at the end of the <ruleSet/> element

src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
ajarmoniuk: will do

src/it/it-compare-dependencies-001/invoker.properties
```diff
@@ -1,2 +1,5 @@
...
+
+invoker.goals.2 = ${project.groupId}:${project.artifactId}:${project.version}:compare-dependencies
+invoker.mavenOpts.2 = -DreportOutputFile=depDiffs-1.txt -DremotePom=localhost:dummy-bom-pom:1.0
```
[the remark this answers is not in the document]
ajarmoniuk: It turns out this test doesn't make much sense here -- we're only producing a comparison report between two poms. I'll take it back from the PR.

ajarmoniuk:
Remarks applied. Please re-review. :)

src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -268,26 +236,60 @@ private static RuleSet loadRuleSet( String serverId, Settings settings, WagonMan
...
+        RuleSet loadedRules = isClasspathUri( rulesUri )
+                ? getRulesFromClasspath( rulesUri, logger )
+                : getRulesViaWagon( rulesUri, logger, serverId, serverId, wagonManager,
+                settings );
+
         RuleSet ruleSet = new RuleSet();
-        boolean rulesUriGiven = isRulesUriNotBlank( rulesUri );
+        ruleSet.setIgnoreVersions( loadedRules.getIgnoreVersions() );
+        ruleSet.setRules( loadedRules.getRules() );
+        return ruleSet;
```
slawekjaranowski, line 247: Why not return loadedRules ... 
missing `ruleSet.setComparisonMethod(...)`

How we can be sure that all fields are cloned, eg when somebody extends `RuleSet` ...

src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -268,26 +236,60 @@ private static RuleSet loadRuleSet( String serverId, Settings settings, WagonMan
...
+    private static RuleSet enrichRuleSet( Collection<String> ignoredVersions, RuleSet originalRuleSet )
+    {
+        RuleSet ruleSet = new RuleSet();
+        if ( originalRuleSet != null )
         {
-            RuleSet loadedRules;
-
-            if ( isClasspathUri( rulesUri ) )
+            ruleSet.setComparisonMethod( originalRuleSet.getComparisonMethod() );
+            ruleSet.setModelEncoding( originalRuleSet.getModelEncoding() );
```
slawekjaranowski, line 271: internal Modello field

src/site/apt/version-rules.apt.vm
```diff
@@ -82,11 +88,88 @@ Version number rules
...
+              <!-- syntax the same as for the general ignoreVersions -->
+              <ignoreVersions>
+                ...
+              </ignoreVersions>
+
+              <!-- optional, comparison method for the rule, one of: maven (default), numeric, mercury -->
+              <comparisonMethod>maven</comparisonMethod>
+
+              <!-- optional; defaults to UTF-8 -->
+              <modelEncoding>UTF-8</modelEncoding>
```
slawekjaranowski, line 147: Internal field not need

src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -1039,4 +1036,164 @@ public PluginPluginUpdatesDetails call()
...
+            }
+            else
+            {
+                instance.ruleSet = isBlank( rulesUri )
+                        ? new RuleSet()
+                        : loadRuleSet( serverId, settings, wagonManager, rulesUri, log );
+            }
+            if ( ignoredVersions != null && !ignoredVersions.isEmpty() )
+            {
+                instance.ruleSet = enrichRuleSet( ignoredVersions, ruleSet );
```
slawekjaranowski, line 1184: Maybe:

```suggestion
                instance.ruleSet = enrichRuleSet( ignoredVersions, instance.ruleSet );
```

src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
ajarmoniuk: Indeed. Not needed at all and the whole method became just one return statement used only in the builder, so I removed it.

ajarmoniuk:
Again, applied all comments. Please review again.

slawekjaranowski (approved):

slawekjaranowski:
@ajarmoniuk Thanks - good job

Issue #610, opened 2022-07-01
Allow setting a project's parent version to a dedicated one


odrotbohm:
Currently, the `update-parent` goal cannot *downgrade* a project's parent version. The documentation lists a `parentVersion` parameter that kind of implies one could set the project's version to a dedicated one. However, the property only controls dependency resolution for the step of changing the version.

bianjp:
I want this feature too.

How about adding a new property named `newVersion` which will be used as the parent version without resolution (works like `newVersion` in `versions:set` goal)? Or adding a new `versions:set-parent` goal?

slawekjaranowski:
`newVersion` property will be ok 

TODO:
 - clarify documentation of `parentVersion` 
 - update examples - https://www.mojohaus.org/versions-maven-plugin/examples/update-parent.html

ajarmoniuk:
I'll take it on my bandwagon.

ajarmoniuk:
@odrotbohm @bianjp it is undocumented and probably a side effect of how available version resolution works, but it is in fact possible to downgrade a non-snapshot version to a specified version. In order to do so, just surround the version in square brackets, like so:

`[0.8.0]`

Then, the version will be *downgraded*. I think I'll add a couple a test documenting/verifying this plus maybe a new argument not to use the square brackets? @slawekjaranowski what do you think?

An explainer in the documentation is a must.

`parentVersion` is a hint to the available resulution system about the *lowest* version from which the sought version range should start -- or can be a version range, in which case the sought version range will be restricted by that version range. And that last possibility is what we're using here.

bianjp:
> it is undocumented and probably a side effect of how available version resolution works, but it is in fact possible to downgrade a non-snapshot version to a specified version.

I know this works, even for snapshot versions. But it still goes through a resolution process and requires that the specified parentVersion already exists in maven repository.

I wish that a new argument could be added to bypass the resolution process.

haj205:
It would really help to skip the resolution process and have a dedicated version updated. In my case, we have different release branches 22.03.100, 22.04.100 and the versions as 22.03.100-SNAPSHOT / 22.04.100-SNAPSHOT and I am not able to update parent versions on different branches with respective release versions.

ajarmoniuk:
Then I think it would be better to reuse `parentVersion` here, but introduce a new parameter, say, `skipResolution`.

[...]

PR #689, opened 2022-09-14
Issue 610 update-parent: add a possibility to skip version resolution and enforce a parent version

ajarmoniuk:
As explained in #610 

If `parentVersion` is set, `skipResolution` will cause the mojo to completely skip version resolution and will simply set the parent version to the given version. 

Added tests + enhanced the documentation.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/UpdateParentMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/UpdateParentMojo.java
@@ -24,6 +24,7 @@ import javax.xml.stream.XMLStreamException;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.metadata.ArtifactMetadataRetrievalException;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
 import org.apache.maven.artifact.versioning.VersionRange;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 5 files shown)


ajarmoniuk:
@slachiewicz @slawekjaranowski please review

src/main/java/org/codehaus/mojo/versions/UpdateParentMojo.java
```diff
@@ -101,60 +117,69 @@ protected void update( ModifiedPomXMLEventReader pom )
...
-        if ( PomHelper.setProjectParentVersion( pom, artifactVersion.toString() ) )
+        ArtifactVersion artifactVersion = findLatestVersion( artifact, targetVersionRange, null,
+                false, allowDowngrade );
+        if ( !shouldApplyUpdate( artifact, getProject().getParent().getVersion(), artifactVersion, forceUpdate ) )
         {
-            getLog().debug( "Made an update from " + currentVersion + " to " + artifactVersion );
-
-            this.getChangeRecorder().recordUpdate( "updateParent", artifact.getGroupId(), artifact.getArtifactId(),
-                                                   currentVersion, artifactVersion.toString() );
+            return null;
```
slawekjaranowski, line 180: Do we have somewhere a log (can be in debug level) why update is not needed? 
It can be useful for debugging.

ajarmoniuk: Actually, yes -- `shouldApplyUpdate` seems to be very chatty. :)

ajarmoniuk:
Could you please rerun failed jobs? Looks like it's infra.

ajarmoniuk:
I don't get it. I think the infra is not well today :)

slachiewicz:
ok, now all green.

PR #693, opened 2022-09-15
Restriction parameter to prevent more than 7 parameters in the next PRs

sultan:
i have a PR waiting to be pushed/suggested but i need these methods to be refactored first so i dont fall into more than 7 parameters compiler errors + the new design i suggest makes the methods more easy to maintain.

i know this does not add any more features but i hope this can be usefull for the quality and ease of maintenance there after.

the basics are sending a Restriction parameter instead of 4 (lower/upper bound and accept booleans), thus gaining 3 more free params to send for other features (next PR to filter RCs is waiting for this one to be accepted)

Thank you for your time

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
@@ -234,18 +234,10 @@ public abstract class AbstractVersionsReport
      * @since 1.0-alpha-1
      */
     protected ArtifactVersion findLatestVersion( Artifact artifact, VersionRange versionRange,
-                                                 boolean allowingSnapshots, boolean usePluginRepositories )
+                                                 Boolean allowingSnapshots, boolean usePluginRepositories )
         throws MavenReportException
     {
-        boolean includeSnapshots = this.allowSnapshots;
-        if ( allowingSnapshots )
-        {
-            includeSnapshots = true;
-        }
-        if ( allowingSnapshots )
-        {
-            includeSnapshots = false;
-        }
+        boolean includeSnapshots = allowingSnapshots != null ? allowingSnapshots : this.allowSnapshots;
         try
         {
             final ArtifactVersions artifactVersions =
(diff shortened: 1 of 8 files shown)


ajarmoniuk:
You can also suppress the checkstyle error by using 

`SuppressWarnings("checkstyle:maxParameterNumber")`

sultan:
Thanks, good to know there are such options !
i find it quite harsh for me to bypass the checkstyle and have like 4 to 5 booleans in the final methods of the next PRs.

I find it easier to digest to cut my PR into two (this one for the refactoring, and another for the exclude milestone/previews feature)

src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
```diff
@@ -234,18 +234,10 @@ protected abstract void doGenerateReport( Locale locale, Sink sink )
...
     {
-        boolean includeSnapshots = this.allowSnapshots;
-        if ( allowingSnapshots )
-        {
-            includeSnapshots = true;
-        }
-        if ( allowingSnapshots )
-        {
-            includeSnapshots = false;
-        }
```
slawekjaranowski, line 248: Interesting ... looks like never was working

slawekjaranowski:
> 
> would it be ok for the team to merge this PR before the next feature, or do you prefer i merge my work into one PR ?

simpler PR easier to review - it is ok for me for splitting

src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
ajarmoniuk: No, not really. I contemplated removing it too, but I saw it being used.

this.allowSnapshots was a `Boolean` value, could also be null. Local `allowingSnaphots` was a primitive `boolean`, but it was a kind of an override for the value in the Mojo.

So, if `allowingSnapshots` method argument was set, the code overrote the value from the Mojo object with the method argument.

sultan: There are two methods with Boolean overrides. One of them was with the boolean primitive and seems not used. I did not felt removing this one. Might consider marking deprecated ?

PR #694, opened 2022-09-16
Issue 342 scm set tag add new elements

ajarmoniuk:
This is some enhancement of the scm-tag functionality. Whereas it was only possible to add new values to the existing scm subelements, this enhancement adds the possibility to add missing elements as well.

In the code, I've replaced stack usage with reflection in PomHelper, but it doesn't really make a difference (can revert it if necessary, I think it's a question of taste). Plus, encapsulated some implementation details, recursion state (drawback of using recursion), and helper methods within a local class.

It's not fully flexible -- it doesn't allow us to create new elements if also the parent element is missing -- but since it's not XSD-aware, it wouldn't know where to put the new element :)


slawekjaranowski (approved):

ajarmoniuk:
I think the PR removing `BaseTestCase` broke this test :laughing: 

I'm gonna create a fixing PR right away

Issue #697, opened 2022-09-16
Maintenance: migrate to JSR330


ajarmoniuk:
Migrate to JSR330. 

Reference: https://github.com/eclipse/sisu.plexus/wiki/Plexus-to-JSR330

ajarmoniuk:
Further steps (for another issue): convert known singletons to JSR330 components (e.g. VersionsHelper).

ajarmoniuk:
@slachiewicz please reopen. The changes haven't been merged.

PR #698, opened 2022-09-16
#697: Migrate to JSR330

ajarmoniuk:
Migrating to JSR330. Notable find: removed the bogus duplicate of the ArtifactResolver component in AbstractVersionsUpdaterMojo (there are two which point to the same component).

Using constructor-style injection, as advised in the reference.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
@@ -19,6 +19,8 @@ package org.codehaus.mojo.versions;
  * under the License.
  */
 
+import javax.inject.Inject;
+
 import java.io.File;
 import java.util.List;
 import java.util.Locale;
@@ -37,7 +39,6 @@ import org.apache.maven.doxia.siterenderer.Renderer;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.reporting.AbstractMavenReport;
@@ -64,17 +65,9 @@ public abstract class AbstractVersionsReport
      *
      * @since 1.0-alpha-3
      */
-    @Component
     protected I18N i18n;
 
-    @Component
     protected RepositorySystem repositorySystem;
-
-    /**
-     * @since 1.0-alpha-3
-     */
-    @Component
-    private ArtifactResolver resolver;
 
     /**
      * Skip entire check.
@@ -89,7 +82,6 @@ public abstract class AbstractVersionsReport
      *
      * @since 1.0-alpha-1
      */
-    @Component
     protected ArtifactMetadataSource artifactMetadataSource;
 
     /**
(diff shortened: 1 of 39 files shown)


ajarmoniuk:
@slachiewicz @slawekjaranowski please review

src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
```diff
@@ -64,16 +65,16 @@
      *
      * @since 1.0-alpha-3
      */
-    @Component
+    @Inject
```
slawekjaranowski, line 68: Why here injecting by fields - not by constructor?

src/main/java/org/codehaus/mojo/versions/CompareDependenciesMojo.java
```diff
@@ -112,11 +116,21 @@
     /**
      * The project builder used to initialize the remote project.
      */
-    @Component
+    @Inject
```
slawekjaranowski, line 119: One by field, others by constructor ...?

src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
ajarmoniuk: Oops, fixing it!

ajarmoniuk: Thanks for spotting this. Corrected.

src/main/java/org/codehaus/mojo/versions/CompareDependenciesMojo.java
ajarmoniuk: Corrected

slachiewicz (approved):

ajarmoniuk:
@slachiewicz why did you close it before merging?

ajarmoniuk:
@slachiewicz @slawekjaranowski please merge ;)

slawekjaranowski:
@ajarmoniuk was merged by direct push in commit: https://github.com/mojohaus/versions-maven-plugin/commit/a54e87fac6597d58eee14ec33c17590524836f48 instead of using GH gui

Another case for manual push is that this PR will be missing from auto generated releases notes. 
@slachiewicz what is purpose for direct push it?

ajarmoniuk:
I see. Thanks for the explanation.

slachiewicz:
this time i was forking from command line, not from GH ui. I've updated, rebased change before push to test it locally.

slachiewicz:
maybe release note action configuration should be updated?

slawekjaranowski:
As I know `release-derafter` only take list of merged PR .. so we should process all important changes by PR with merged status ... 😄

slachiewicz:
i've updated release note with my merge PR plus few small styling updates

slawekjaranowski:
We will see what will happen after next release note rebuilds ... try to re-run Release Drafter action

slachiewicz:
all my changes dropped :( thx for explanation

ajarmoniuk:
@slachiewicz I've noticed that you're now trying to replace constructor-based injection with injecting fields. -- that is, placing the @Inject annotations next to fields. While this is certainly possible, it's not advised to do so as it's using reflection. 

Please see https://github.com/eclipse/sisu.plexus/wiki/Plexus-to-JSR330#Basics

slachiewicz:
Yes, i expect that without constructors it will be easier to refactor code to not use deprecated classes from Maven 2.
Someone can migrate Mojo one by one.
Later we can introduce constructors with new proper parameters.

PR #701, opened 2022-09-17
Fixing #700: Adapting SetScmTagMojoTest to work without BaseMojoTestCase

ajarmoniuk:

PR #702, opened 2022-09-17
Refactoring UpdateScope, VersionDetails and related classes

ajarmoniuk:
The goal of this exercise is to simply VersionDetails and related classes. I've noticed that a lot of logic related to version management is spread across multiple classes. In particular, UpdateScope which is supposed to be a simple enum-like class and a container for the segment, has grown to some utility class with code duplication. Cleaned that up.

The ultimate goal is to replace version management helpers with JSR330 components, also to split VersionComparator classes from their utility part (segment manipulation) -- but that's for another PR. I want to keep the scope small so that it doesn't become another gargantuan PR to review :laughing: 

*Please note: until #701 is merged in, there are two commits, then I'll merge and squash.*


sultan:
somehow i cant make my ide checkout the pr, i think it will be easier for both of us if you can merge your PR first, then i'll have a nice branch to checkout and merge on my side. would it be ok for you ?

slachiewicz (approved):

Issue #704, opened 2022-09-18
Remove usages of deprecated Maven 2 references (like maven-compat)


slachiewicz:
we should migrate all to new Maven 3 replacements

MavenProjectBuilder projectBuilder,
ArtifactMetadataSource artifactMetadataSource,
WagonManager wagonManager,
 ArtifactResolver artifactResolver

slachiewicz:
#559

slawekjaranowski:
Next one: #713

ajarmoniuk:
Working on migrating to Maven 3.

slawekjaranowski:
Please one by one if possible 😄

ajarmoniuk:
@slawekjaranowski I have a question.

Do we still need to provide Maven 2 backward compatibility in case of plugin repos? I've created this, but am now thinking whether it is really needed...

```java
/**
 * Provides backward-compat with 2.x that allowed plugins like the maven-remote-resources-plugin:1.0 to
 * populate the builder configuration with model repositories instead of artifact repositories.
 *
 * @param repositories list of remote repositories
 * @param session {@link RepositorySystemSession} instance
 * @param repositorySystem {@link RepositorySystem} instance
 * @return list of repositories converted to {@link ArtifactRepository} objects
 * @throws InvalidRepositoryException if the remote repository cannot be reconstructed
 *
 * @since 2.14.0
 */
public static List<ArtifactRepository> normalizeToArtifactRepositories( List<?> repositories,
                                                                        RepositorySystemSession session,
                                                                        RepositorySystem repositorySystem )
        throws InvalidRepositoryException
{
    try
    {
        return repositories
                .stream()
                .filter( o -> o instanceof Repository )
                .map( o -> (Repository) o )
                .map( r ->
                {
                    try
                    {
                        ArtifactRepository repo = repositorySystem.buildArtifactRepository( (Repository) r );
                        repositorySystem.injectMirror( session, singletonList( repo ) );
                        repositorySystem.injectProxy( session, singletonList( repo ) );
                        repositorySystem.injectAuthentication( session, singletonList( repo ) );
                        return repo;
                    }
                    catch ( InvalidRepositoryException e )
                    {
                        throw new IllegalArgumentException( e );
                    }
                } )
                .collect( Collectors.toList() );
    }
    catch ( IllegalArgumentException e )
    {
        throw (InvalidRepositoryException) e.getCause();
    }
}
```

slawekjaranowski:
[...]

PR #710, opened 2022-09-18
In a Mojo we can directly use @Inject - Sisu will do all work for us

slachiewicz:


slawekjaranowski (approved):
Why inject by fields are better than inject by constructors? 😄

slachiewicz:
It's worse ofc. I was thinking that this can help doing refactor for #704 (in smaller parts (due to lack of time and maybe smaller pr). Now after some time i don't think this would be really useful

slawekjaranowski:
>.... Now after some time i don't think this would be really useful

So do you want to process it or we can close?

Issue #717, opened 2022-09-21
Plugin uses the reserved maven prefix "maven."


ajarmoniuk:
As noted in #454, versions-maven-plugin should not be using the `maven` property prefix. Hereby changing this to `mojohaus.versions`, as agreed there. 

This is a breaking change.

pzygielo:
- https://github.com/mojohaus/versions-maven-plugin/pull/718

pzygielo:
- https://github.com/mojohaus/versions-maven-plugin/pull/724

daniel-beck:
Probably more accurate to call it _reserved_?

slawekjaranowski:
For me `versions.` as prefix will be ok.

As such changes are not compatible, I would see one property per one PR with title like `Rename property xxx to yyy`

I would to show those change in clear way in release notes.

ajarmoniuk:
> I would to show those change in clear way in release notes.

Should I update something?

slawekjaranowski:
I would like to review it and discus more ....

I agree that properties begins from `maven.` should be not used.

What I see:
- in project most of properties are not prefixed  
- if prefix is used it is `versions` - not `version` ( at most case ) 

To think about:
- whether to use a prefix at all
- if property is dedicated for cli maybe without prefix is better ... i'm layzy
- whether all `Mojo Parameters` need property?

ajarmoniuk:
My thoughts

- prefixes makes sense if the plugin is used in combination with other plugins
- currently this is difficult to imagine, but let's say we introduce a feature that would work in combination with the enforcer plugin, i.e. conditions for failure
- I don't think all pom.xml properties need a user property, but it makes it much easier to test the plugin using options passed via invoker.properties
- long property names make the possibility of collisions (with other plugins) less likely, but they can of course still occur
- if property is dedicated for cli maybe without prefix is better ... i'm layzy -> let's just use 'versions' for short/collision prone property names

PR #718, opened 2022-09-21
#717: Changing the property prefix to mojohaus.versions

ajarmoniuk:


slawekjaranowski:
I'm not sure if we should brake compatibility for old properties in simply way.

We rename two old properties:
 - maven.version.rules.serverId
 - maven.version.rules
 
Probably such properties not widely used, but in this case I would prefer one PR for one property to show such change in releases notes.

ajarmoniuk:
Yeah, that was my concern as well... And there's no easy way of providing an alias to a property, besides creating a second parameter.

Should we just leave these two parameters as they are and just rename the third one?

slawekjaranowski:
I don't expect a big impacts on users - I think that when we show it in release notes will be acceptable - parameter name for configuration are not changed

ajarmoniuk:
I've just noticed this:

```
@Parameter( property = "versions.skip" )
```

Should we go for the `versions.` prefix then?

pzygielo:
> Should we go for the `versions.` prefix then?

:+1:

pzygielo:
Just linking to
- #717

Issue #265, opened 2018-03-13
Different profile used for versions:set and versions:revert .


garretwilson:
I have an aggregate POM situation that does _not_ use inheritance, i.e the aggregated POMs do not have a parent-child relationship with the aggregating POM. The directory structure looks like this:

* `aggregate/`
* `foo/`
* `bar/`

The aggregating POM in the `aggregate/` directory specifies the `foo` module:

```xml
	<modules>
		<module>../foo</module>
```

It also has a profile `foobar` which brings in the `bar` module:

```xml
	<profiles>
		<profile>
			<id>foobar</id>
			<modules>
				<module>../bar</module>
			</modules>
		</profile>
```

Now here's the interesting part. If I go into the `aggregate/` directory and issue the following command:

```
mvn versions:set -DnewVersion=1.2.3 -DprocessAllModules
```

[Thanks to the `processAllModules` option](https://stackoverflow.com/a/49246337/421049), this updates the aggregate POM version and the aggregated module versions both (both `foo` and `bar`, as well as there interdependent dependency references)! Yay!

But if I then try to revert the changes:

```
mvn versions:revert -DprocessAllModules
```

This command does not revert the POM for the `bar` module!!! I have to specify the `foobar` profile to get it to be reverted too:

```
mvn versions:revert -DprocessAllModules -P foobar
```

One can make arguments either way about whether the plugin should automatically process modules that are part of profiles that weren't indicated, and for the purposes of this ticket it really doesn't matter.

The important thing is that: shouldn't both goals `versions:set` and `versions:revert` use the same module inclusion rules for the `processAllModules` option? Put another way, shouldn't I be able to use the same options I used for changing the version to commit or revert that version? I don't understand what different semantics of `processAllModules` would require me to indicate a profile for one of the goals but not the other.

Pirayya:
I get the exact same issue in one of my projects (writing this comment just to "bump" the issue). I think `versions:revert` and `versions:set` should have the same behaviour and alter the same modules!

ajarmoniuk:
Looking into it. I know it's an old one.

ajarmoniuk:
RevertMojo only recovers the old copy of the root pom.xml from the backup in the .versionsBackup file. This could indeed be enhanced.

PR #719, opened 2022-09-22
Issue #265: revert mojo to scan child modules instead of relying on maven

ajarmoniuk:
Using `PomHelper::getAllChildModules` to retrieve all modules in the reactor to get files to revert. Relying on maven to process all enabled profiles did not revert all files changed by `SetMojo`, because `SetMojo` also uses `getAllChildModules`.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/RevertMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/RevertMojo.java
@@ -19,16 +19,25 @@ package org.codehaus.mojo.versions;
  * under the License.
  */
 
-import java.io.File;
-import java.io.IOException;
+import javax.inject.Inject;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Set;
+
+import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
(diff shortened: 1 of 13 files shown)


ajarmoniuk:
@slachiewicz @slawekjaranowski please review and merge

src/main/java/org/codehaus/mojo/versions/RevertMojo.java
```diff
@@ -48,24 +56,62 @@
...
-    public void execute()
-        throws MojoExecutionException, MojoFailureException
+    /**
+     * Whether to start processing at the local aggregation root (which might be a parent module
+     * of that module where Maven is executed in, and the version change may affect parent and sibling modules).
+     * Setting to false makes sure only the module (and it's submodules) where Maven is executed for is affected.
+     *
+     * @since 2.13.0
+     */
+    @Parameter( property = "processFromLocalAggregationRoot", defaultValue = "true" )
```
slawekjaranowski, line 66: Why not the same name as for set goal - `processAllModules`?

src/main/java/org/codehaus/mojo/versions/RevertMojo.java
```diff
@@ -48,24 +56,62 @@
...
+    protected RevertMojo( MavenProjectBuilder projectBuilder )
+    {
+        this.projectBuilder = projectBuilder;
+    }
+
+    public void execute() throws MojoExecutionException, MojoFailureException
     {
-        File outFile = project.getFile();
-        File backupFile = new File( outFile.getParentFile(), outFile.getName() + ".versionsBackup" );
+        final MavenProject project = !processFromLocalAggregationRoot
```
slawekjaranowski, line 82: Please don't override field defined in class

src/main/java/org/codehaus/mojo/versions/RevertMojo.java
ajarmoniuk: Because it's an analogue of `processFromLocalAggregationRoot` from `SetMojo`, that's why it has the same name as that property :)

Issue #343, opened 2019-03-30
Report parent updates


reallyinsane:
Currently there is no way to generate a report including the update information about the parent. There is a report for dependency updates (dependency-updates-report) and one for plugin updates (plugin-updates-report). I'd like to have the information about the parent too.

There could be two ways to achive this:

- create a new goal for generating a report about the parent (parent-updates-report)
- create a single report for everything (parent, dependencies, plugins, properties) with options to disable parts of them

What do you think?

P.S. I need a report because I use it to generate metrics about the possible updates for a project. By now the information about the parent is missing.

helpermethod:
So much this!

A complete report would be very helpful, as I would like to generate pull requests based on the report.

ajarmoniuk:
Hi there. Do you still need this feature @reallyinsane ?

reallyinsane:
Hi,
no offense, but the issue is three years old. I'd really like to have this feature. If you don't have time for it, I could also do the changes on my own and create a pull request.

Regards

Matthias

ajarmoniuk:
I've only came to this project a few weeks ago @reallyinsane

ajarmoniuk:
@reallyinsane let me know if you're willing to, otherwise I'll make one. Btw, aggregate support will be added as a different item.

reallyinsane:
> @reallyinsane let me know if you're willing to, otherwise I'll make one. Btw, aggregate support will be added as a different item -- #333

Sorry for the late reply. Now all is implemented already. Thank you for that.

PR #722, opened 2022-09-23
#343 Parent updates report

ajarmoniuk:
Implementing the requested enhancement + added an it and a unit test.

I plan on adding aggregate reports in another PR as a response to a request in another issue -- I think it's a perfectly valid request.

Also, I want to refactor the whole reports suite (and maybe try using velocity?), all of them, but that's for another PR.


ajarmoniuk:
@slawekjaranowski Please review

PR #724, opened 2022-09-27
#717: Changing the reserved prefix to "versions"

ajarmoniuk:
Alternative to #718, changing the prefix to "versions" since it's already being used for `versions.skip`.


pzygielo:
- https://github.com/mojohaus/versions-maven-plugin/issues/717

Issue #454, opened 2021-03-21
Unexpected behavior of -DallowMajorReleases=false


daniel-beck:
Recently seen output of `display-dependency-updates`:

```
$ mvn org.codehaus.mojo:versions-maven-plugin:2.8.1:display-dependency-updates
[INFO] Scanning for projects...
[INFO] 
[INFO] -------------------< org.jenkins-ci:update-center2 >--------------------
[INFO] Building Jenkins Update Center Generator 3.6-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- versions-maven-plugin:2.8.1:display-dependency-updates (default-cli) @ update-center2 ---
[INFO] The following dependencies in Dependency Management have newer versions:
[INFO]   com.github.spotbugs:spotbugs-annotations ............. 3.1.12 -> 4.2.2
[INFO]   junit:junit ........................................... 4.13 -> 4.13.2
[INFO]   org.codehaus.mojo:animal-sniffer-annotations ............ 1.18 -> 1.20
[INFO] 
[INFO] The following dependencies in Dependencies have newer versions:
[INFO]   com.squareup.okhttp3:mockwebserver ............ 4.8.0 -> 5.0.0-alpha.2
[INFO]   com.squareup.okhttp3:okhttp-urlconnection ..... 4.8.0 -> 5.0.0-alpha.2
[INFO]   commons-codec:commons-codec .................. 1.14 -> 20041127.091804
[INFO]   commons-io:commons-io ......................... 2.7 -> 20030203.000550
[INFO]   jaxen:jaxen ............................... 1.2.0 -> 1.2.0-atlassian-2
[INFO]   org.jetbrains.kotlin:kotlin-stdlib-common ......... 1.3.72 -> 1.5.0-M1
[INFO] 
[INFO] The following dependencies in Plugin Dependencies have newer versions:
[INFO]   org.codehaus.mojo:extra-enforcer-rules .................... 1.2 -> 1.3
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.058 s
[INFO] Finished at: 2021-03-20T10:59:07+01:00
[INFO] ------------------------------------------------------------------------
```

[...]

PR #733, opened 2022-10-02
#454: Introducing "infinity-like" range calculus for dependency range computation

ajarmoniuk:
Instead of using the next version for range calculation when considering version upgrades, I've devised a sort of a bounded artifact version where we use the concept of "infinity" on a given segment position to match all possible upper versions on the position, but nothing from the more major segment. So, 2.0.0-beta or similar won't be matched anymore when considering minor updates of the 1.x.x branch.

Tests added as unit tests; in particular, two for testing the case explained in #454 in `DisplayDependencyUpdatesMojoTest`

Some minor refactoring and typo corrections.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReportRenderer.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReportRenderer.java
@@ -780,40 +780,47 @@ public abstract class AbstractVersionsReportRenderer
 
     protected String getLabel( ArtifactVersion version, AbstractVersionDetails versions )
     {
-        String label = null;
+        if ( equals( version, versions.getNewestUpdate( of( SUBINCREMENTAL ) ) ) )
+        {
+            return getText( "report.latestSubIncremental" );
+        }
+
+        if ( equals( version, versions.getOldestUpdate( of( SUBINCREMENTAL ) ) ) )
+        {
+            return getText( "report.nextVersion" );
+        }
+
+        if ( equals( version, versions.getOldestUpdate( of( INCREMENTAL ) ) ) )
+        {
+            return getText( "report.nextIncremental" );
+        }
+
+        if ( equals( version, versions.getNewestUpdate( of( INCREMENTAL ) ) ) )
+        {
+            return getText( "report.latestIncremental" );
+        }
+
+        if ( equals( version, versions.getOldestUpdate( of( MINOR ) ) ) )
+        {
+            return getText( "report.nextMinor" );
+        }
+
+        if ( equals( version, versions.getNewestUpdate( of( MINOR ) ) ) )
+        {
+            return getText( "report.latestMinor" );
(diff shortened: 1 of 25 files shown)


ajarmoniuk:

src/main/java/org/codehaus/mojo/versions/AbstractVersionsReportRenderer.java
```diff
@@ -780,40 +780,23 @@ private Set<String> getVersionsInRange( Property property, PropertyVersions vers
...
+                            ? getText( "report.latestIncremental" )
+                            : equals( version, versions.getOldestUpdate( of( MINOR ) ) )
+                                ? getText( "report.nextMinor" )
+                                : equals( version, versions.getNewestUpdate( of( MINOR ) ) )
+                                    ? getText( "report.latestMinor" )
+                                    : equals( version, versions.getOldestUpdate( of( MAJOR ) ) )
+                                        ? getText( "report.nextMajor" )
+                                        : equals( version, versions.getNewestUpdate( of( MAJOR ) ) )
+                                            ? getText( "report.latestMajor" )
+                                            : "";
```
slawekjaranowski, line 799: `if/else` was more readable for me ... but it can be my personal feeling

src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -691,12 +691,19 @@ private DependencyManagement getProjectDependencyManagement( MavenProject projec
...
+                    + "allowIncrementalUpdates must be true" );
+        }
+
+        return allowAnyUpdates && allowMajorUpdates && allowMinorUpdates
+                ? empty()
+                : allowMajorUpdates && allowMinorUpdates
+                    ? of( MAJOR )
+                    : allowMinorUpdates
+                        ? of( MINOR )
+                        : of( INCREMENTAL );
```
slawekjaranowski, line 706: difficult to reading ... maybe simple if/else 😄

src/main/java/org/codehaus/mojo/versions/api/Segment.java
```diff
@@ -58,6 +51,16 @@ public static Segment of( int index )
         }
```
slawekjaranowski, line 51: line 43 - 51
```
values()[index]
```
?

src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
ajarmoniuk: Okay, I prefer the Elvis operator, but I'll change it to the if ladder :)

ajarmoniuk:
Changes applied. Please re-review :)

slawekjaranowski (approved):

Issue #729, opened 2022-09-30
Flipping it-property-updates-report-002-slow


ajarmoniuk:
This exercise is about investigating the flipper mentioned in the summary. As far as I've seen, it only occurs on MacOS.

ajarmoniuk:
The test doesn't take into account that the overall execution time may be longer due to artifact retrieval. Strangely enough, this happens almost exclusively on MacOS platforms. Anyway, I think this should be done differently, e.g. by profiling and not using an it test.

ajarmoniuk:
Rebased

ajarmoniuk:
@slachiewicz please close

Thanks

PR #735, opened 2022-10-03
#729: Ignoring the flipper

ajarmoniuk:
This test should be used for profiling and not as part of integration testing. Caching rule resolution is a good start, but the actual performance depends on the machine and conditions of the cluster, but tests should not really depend on the volatile cluster state.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -368,6 +368,16 @@
           <pomIncludes>
             <pomInclude>*/pom.xml</pomInclude>
           </pomIncludes>
+          <pomExcludes>
+            <!--
+              ! This test is not really intended as part of regular integration test suite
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -368,6 +368,9 @@
           <pomIncludes>
             <pomInclude>*/pom.xml</pomInclude>
           </pomIncludes>
+          <pomExcludes>
+            <pomExclude>it-property-updates-report-002-slow/*</pomExclude>
+          </pomExcludes>
```
slawekjaranowski, line 373: please add a short comment why is excluded ...

ajarmoniuk:
Added a more elaborate description.

slawekjaranowski (approved):

Issue #499, opened 2021-09-28
update-parent does not respect -DallowMajorUpdates


askoog:
The  `update-parent` goal does not respect the properties `allowMajorUpdates`, `allowMinorUpdates` and `allowIncrementalUpdates` that `use-latest-version` goal uses. This is somewhat confusing (even though the documentation states that the property is not used)

Example pom file
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"     
        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">  
        <modelVersion>4.0.0</modelVersion>                                                                   
                                                                                                             
        <parent>                                                                                             
            <groupId>org.springframework.boot</groupId>                                                      
            <artifactId>spring-boot-starter-parent</artifactId>                                              
            <version>1.5.22.RELEASE</version>                                                                
        </parent>                                                                                            
                                                                                                             
        <artifactId>test-pom</artifactId>                                                                    
</project>                                                                                                   
```

Updating the parent version will update to a new major version even if `allowMajorUpdates` is declared

```
$ mvn versions:update-parent -DallowMajorUpdates=false
...
[INFO] -----------------< org.springframework.boot:test-pom >------------------
[INFO] Building test-pom 1.5.22.RELEASE
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- versions-maven-plugin:2.2:update-parent (default-cli) @ test-pom ---
[INFO] Updating parent from 1.5.22.RELEASE to 2.5.1
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

Since, for example the java version used and other breaking changes might be inherited from the parent, there should be a way to update the parent without updating to a new major version

PR #737, opened 2022-10-05
#499: Add segment selection to update parent

ajarmoniuk:
This feature has been requested in #499. Applying segment selection to `UpdateParentMojo`.

Tests realised as unit tests in `UpdateParentMojoTest`

Converted `it-update-parent-002` and `it-update-parent-003` to unit tests.


src/main/java/org/codehaus/mojo/versions/UpdateParentMojo.java
```diff
@@ -95,6 +103,30 @@ public class UpdateParentMojo extends AbstractVersionsUpdaterMojo
                 defaultValue = "false" )
     protected boolean allowDowngrade;
 
+    /**
+     * Whether to allow the major version number to be changed.
+     *
+     * @since 1.2
+     */
+    @Parameter( property = "allowMajorUpdates", defaultValue = "true" )
+    protected boolean allowMajorUpdates = true;
```
slawekjaranowski, line 112: It is new option for this goal ... so `since` should be `2.13.0`

similar for rest new parameters.

Issue #637, opened 2022-08-18
{Dependency,Plugin}UpdatesReport overhaul


ajarmoniuk:
I see that there's quite some code duplication between DependecyUpdatesReport and PluginUpdatesReport. The classes could possibly use some refactoring. Plus maybe a name change to be in line with the convention (name should probably end with Mojo).

Speaking of which, DependencyUpdatesRenderer uses a completely different interface than DependencyUpdatesXmlRenderer. There's no common factory or a common interface to use the two and the user classes, need to create and handle the two different classes differently.

Here we should create a common API with probably a common abstract factory to instantiate the correct implementation of the renderer.

I appreciate this is not a functional nor a performance issue, but rather a code quality one. I will gladly take it.

ajarmoniuk:
Heads up: it looks like using dependency inversion can offer much here, going with it.

ajarmoniuk:
Extra item to do: convert xml report generators to "real" `ReportGenerator`s using e.g. the `RandomAccessSink` or the normal text sink. Then it would be possible to generate those XML reports from `<reporting/>` too. The `execute()` invocation would be slightly different, it would simply delegate the generation to two different renderers (with their two different sinks).

ajarmoniuk:
Also, this is probably wrong:

If a dependency is in both management and normal dependencies, the one from management will get overwritten.

```
allUpdates = new TreeMap<>( new DependencyComparator() );
allUpdates.putAll( model.getDependencyManagementUpdates() );
allUpdates.putAll( model.getDependencyUpdates() );
```

ajarmoniuk:
Still work in progress, so far have introduced a common model for the renderers and changed renderers into injectable components. It's not done yet -- I'd like to use a templating engine like Velocity. Sadly, it's not possible to inject a "context" into AptParser, otherwise I'd use it and would create a few macros to process content. 

For now, refactored and unified the renderers.

https://github.com/mojohaus/versions-maven-plugin/compare/master...ajarmoniuk:versions-maven-plugin:doxia-renderers

ajarmoniuk:
Please close @slawekjaranowski @slachiewicz

PR #738, opened 2022-10-06
#637: Refactoring report renderers

ajarmoniuk:
Trying to abstract a common model and make report renders more cohesive.

There is still way of improvement (for example: the report renderers should *not* compute things (extract data from VersionsHelper) on themselves, but they should just present data) but due to the size of this PR I've decided not to drag this indefinitely, but make a stop already on this stage. I think this could improve things already.


src/main/java/org/codehaus/mojo/versions/reporting/DependencyUpdatesReportRenderer.java
```diff
@@ -0,0 +1,134 @@
...
+import static org.codehaus.mojo.versions.api.Segment.INCREMENTAL;
+import static org.codehaus.mojo.versions.api.Segment.MAJOR;
+import static org.codehaus.mojo.versions.api.Segment.MINOR;
+import static org.codehaus.mojo.versions.api.Segment.SUBINCREMENTAL;
+
+/**
+ * @param <K> type of the model
+ * @since 1.0-beta-1
+ */
+@Contract( threading = SAFE )
```
slawekjaranowski, line 44: ? what is it? from `org.apache.http.annotation.Contract`

src/main/java/org/codehaus/mojo/versions/reporting/ParentUpdatesReportRenderer.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.maven.doxia.sink.Sink;
+import org.codehaus.mojo.versions.reporting.model.ParentUpdatesModel;
+import org.codehaus.plexus.i18n.I18N;
+
+import static org.apache.http.annotation.ThreadingBehavior.SAFE;
+
+/**
+ * @since 1.0-beta-1
+ */
+@Contract( threading = SAFE )
```
slawekjaranowski, line 34: ?

src/main/java/org/codehaus/mojo/versions/reporting/PluginUpdatesReportRenderer.java
```diff
@@ -45,228 +46,160 @@
 /**
  * @since 1.0-beta-1
  */
-public class PluginUpdatesRenderer
-    extends AbstractVersionsReportRenderer
+@Contract( threading = SAFE )
```
slawekjaranowski, line 49: ?

src/main/java/org/codehaus/mojo/versions/reporting/PropertyUpdatesReportRenderer.java
```diff
@@ -0,0 +1,434 @@
...
+import static java.util.Optional.empty;
+import static java.util.Optional.of;
+import static org.apache.http.annotation.ThreadingBehavior.SAFE;
+import static org.codehaus.mojo.versions.api.Segment.INCREMENTAL;
+import static org.codehaus.mojo.versions.api.Segment.MAJOR;
+import static org.codehaus.mojo.versions.api.Segment.MINOR;
+import static org.codehaus.mojo.versions.api.Segment.SUBINCREMENTAL;
+
+/**
+ * @since 1.0-beta-1
```
slawekjaranowski, line 51: new or copied class?

src/main/java/org/codehaus/mojo/versions/reporting/ReportRendererFactoryImpl.java
```diff
@@ -0,0 +1,83 @@
...
+import org.codehaus.mojo.versions.reporting.model.DependencyUpdatesModel;
+import org.codehaus.mojo.versions.reporting.model.ParentUpdatesModel;
+import org.codehaus.mojo.versions.reporting.model.PluginUpdatesModel;
+import org.codehaus.mojo.versions.reporting.model.PropertyUpdatesModel;
+import org.codehaus.plexus.i18n.I18N;
+
+/**
+ * Factory for report renderers
+ */
+@Named
```
slawekjaranowski, line 38: Please add `@Singleton` - it looks as stateless

src/main/java/org/codehaus/mojo/versions/xml/PluginUpdatesXmlRenderer.java
```diff
@@ -9,7 +9,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE-2.0
```
slawekjaranowski, line 12: not related change, 
by the way license headers for Mojo House project should be discussed on dev list.
I'm not sure if should be ... `Licensed to the Apache Software Foundation (ASF)`

I would like to not touch this now.

src/main/java/org/codehaus/mojo/versions/reporting/PropertyUpdatesReportRenderer.java
ajarmoniuk: See https://github.com/mojohaus/versions-maven-plugin/blob/master/src/main/java/org/codehaus/mojo/versions/PropertyUpdatesRenderer.java#L38

src/it/it-abstract-versions-report-001/verify.groovy
```diff
@@ -1,15 +1,17 @@
...
-
-assert dependencyUpdatesReport.exists(  )
+dependencyUpdatesReport = new File( basedir, "target/site/dependency-updates-report.html" ).text
+        .replaceAll( '<[^>]+>', ' ' )
+        .replaceAll( '&[^;]+;', ' ' )
+        .replaceAll( '\\s+', ' ' )
 
 // some basic (=not comprehensive) checks on the contents of the dependency report
-assert dependencyUpdatesReport.text =~ /\b1\.1\.0-2\b/  // current version of the dependency
-assert dependencyUpdatesReport.text =~ /\b3\.0\b/       // latest major available version
```
slawekjaranowski, line 7: I start thinking that we can use external library for parsing html like https://jsoup.org/ ...
But such improvement is for another PR 😄

src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReportMojo.java
```diff
@@ -52,8 +55,8 @@
  * @since 1.0-beta-1
  */
 @Mojo( name = "dependency-updates-report",
-       requiresDependencyResolution = ResolutionScope.RUNTIME, threadSafe = true )
-public class DependencyUpdatesReport extends AbstractVersionsReport
+       requiresDependencyResolution = ResolutionScope.RUNTIME, threadSafe = false )
```
slawekjaranowski, line 58: Why change `threadSafe` true -> false?

src/main/java/org/codehaus/mojo/versions/PluginUpdatesReportMojo.java
```diff
@@ -50,8 +53,8 @@
  * @since 1.0-beta-1
  */
 @Mojo( name = "plugin-updates-report", requiresDependencyResolution = ResolutionScope.RUNTIME,
-       threadSafe = true )
-public class PluginUpdatesReport extends AbstractVersionsReport
+       threadSafe = false )
+public class PluginUpdatesReportMojo extends AbstractVersionsReport<PluginUpdatesModel>
```
slawekjaranowski, line 57: `threadSafe` true -> false?

src/main/java/org/codehaus/mojo/versions/PropertyUpdatesReportMojo.java
```diff
@@ -47,11 +49,9 @@
  * @since 1.0-beta-1
  */
 @Mojo( name = "property-updates-report", requiresDependencyResolution = ResolutionScope.RUNTIME,
-       threadSafe = true )
-public class PropertyUpdatesReport
-    extends AbstractVersionsReport
+       threadSafe = false )
```
slawekjaranowski, line 52: threadSafe?

src/it/it-abstract-versions-report-001/verify.groovy
ajarmoniuk: Yes and no. This is faster and more concise, but on the other hand, it is still using regexp which is SLOW. However, I only did it so that, if the test is wrong, it's much easier to see the differences. :)

src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReportMojo.java
ajarmoniuk: It was set wrong in the original class to begin with. Multiple threads can share access to the same sink object.

src/main/java/org/codehaus/mojo/versions/PluginUpdatesReportMojo.java
ajarmoniuk: Same as the previous answer

src/main/java/org/codehaus/mojo/versions/PropertyUpdatesReportMojo.java
ajarmoniuk: Same as before

src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReportMojo.java
slawekjaranowski: new Sink object is create for each plugin executions ...
https://github.com/apache/maven-reporting-impl/blob/maven-reporting-impl-3.2.0/src/main/java/org/apache/maven/reporting/AbstractMavenReport.java#L164

IMHO it is not shred by multiple thread ... maybe I don't see something ...

ajarmoniuk: Okay, maybe I was wrong then. 

On the other hand, this needs to be revised. It doesn't fail multithreaded tests without a reason.

ajarmoniuk: Updated

Issue #725, opened 2022-09-28
Migrate site to markdown?


ajarmoniuk:
Just an idea. Should we migrate the site to markdown? Easier to maintain imho, plus it's compatible with github and many tools including builtin plugin in IntelliJ

sultan:
other projects like TomEE use asciidoc, but Markdown is a bit simpler to use and generate html. no dependencies on jbake or the kind. difficulty arised trying migrating all the pages... and not loose content.

maintaining the docs with any LML has been a breeze for me these past months.
MD has almost no entry cost for accustomed github users.

slawekjaranowski:
We can do it - please one document per one PR.

By the way content should be verified if is still valid and corresponding to current implementations ... 😄 

- `.vm` files is only needed if we have properties in document

- instead of `#` for headers - we can use  
```
Header level 1
===========

Header level 2 
--------------
```

pzygielo:
- https://github.com/mojohaus/versions-maven-plugin/pull/734

ajarmoniuk:
@slawekjaranowski @slachiewicz I think this can be closed 👍

PR #739, opened 2022-10-06
#725 Migrating usage.apt.vm to Markdown

ajarmoniuk:
Migrating usage.apt.vm to Markdown


src/site/markdown/usage.md.vm
```diff
@@ -0,0 +1,259 @@
...
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+## Velocity uses # as a prefix for macros and ## as a prefix for comments
+#set($h1 = '#')
+#set($h2 = '##')
+#set($h3 = '###')
+#set($h4 = '####')
```
slawekjaranowski, line 26: not used

ajarmoniuk: Yes, it's being used, see e.g.

https://github.com/ajarmoniuk/versions-maven-plugin/blob/273bb8ba8f3f6ce88126df4f501a12d247e89221/src/site/markdown/usage.md.vm#L28

https://github.com/ajarmoniuk/versions-maven-plugin/blob/273bb8ba8f3f6ce88126df4f501a12d247e89221/src/site/markdown/usage.md.vm#L33

https://github.com/ajarmoniuk/versions-maven-plugin/blob/273bb8ba8f3f6ce88126df4f501a12d247e89221/src/site/markdown/usage.md.vm#L69

etc :)

slawekjaranowski: I see, only `h4` is not used 😄

ajarmoniuk: Removed h4 :)

It's a pity there's h3 though, otherwise I could have used ---- and ===.

slawekjaranowski: Another idea for it https://issues.apache.org/jira/browse/DOXIA-677

ajarmoniuk: Regarding Doxia.

While working on those report renderers, I thought that it would actually be much better to be able to parse a velocity macro with some context injected, and simply render that context. Now we're doing that from the renderer class injecting some primitives into the sink, which "prints" everything using velocity, but much of velocity's potential goes to waste this way.

slawekjaranowski:
please rebase with the latest master version and squash to final commit

ajarmoniuk:
Should be good now.

PR #745, opened 2022-10-08
#283: allowMinorUpdates false should imply allowMajorUpdates false

ajarmoniuk:
Implementing #283: allowMinorUpdates false should also imply allowMajorUpdates equal to false.

Likewise, allowIncrementalUpdates false implies allowMinorUpdates and allowMajorUpdates false.

Not enforced, but the values are assumed. I think the changes are also intuitive, i.e. what most users expect.

Documentation has been updated.


ajarmoniuk:
Please review @slawekjaranowski

slawekjaranowski (approved):

PR #749, opened 2022-10-10
Caching update information in reports to increase performance

ajarmoniuk:
Noticed some duplication of calls to `AbstractVersionDetails::getOldestUpdate` and similar: e.g. the same information is retrieved first during the rendering of the "Overview" section to gather aggregate information, then the call is repeated to render update details.


ajarmoniuk:

slawekjaranowski (approved):

Issue #755, opened 2022-10-13
Regression: XML reports are serialized incorrectly


ajarmoniuk:
While looking at XML renderers, I've noticed that they're now rendered incorrectly. Please hold the release. I'm fixing it.
```xml
<PluginUpdatesReport>
	<summary>
		<usingLastVersion>1</usingLastVersion>
		<nextVersionAvailable>1</nextVersionAvailable>
		<nextIncrementalAvailable>0</nextIncrementalAvailable>
		<nextMinorAvailable>10</nextMinorAvailable>
		<nextMajorAvailable>1</nextMajorAvailable>
	</summary>
	<pluginManagements>
		<pluginManagement>
			<groupId>org.apache.maven.plugins</groupId>
			<artifactId>maven-antrun-plugin</artifactId>
			<currentVersion>1.3</currentVersion>
			<nextVersion>1.4</nextVersion>
			<minors>
				<optional[minor]>1.4</optional[minor]>
				<optional[minor]>1.5</optional[minor]>
				<optional[minor]>1.6</optional[minor]>
				<optional[minor]>1.7</optional[minor]>
				<optional[minor]>1.8</optional[minor]>
			</minors>
			<majors>
				<optional[major]>3.0.0</optional[major]>
				<optional[major]>3.1.0</optional[major]>
			</majors>
			<status>minor available</status>
		</pluginManagement>
...
```

PR #756, opened 2022-10-13
Fixing #755: Segment names containing Optional were used as tag names in XML reports

ajarmoniuk:


ajarmoniuk:
@slawekjaranowski @slachiewicz please review

ajarmoniuk:
Alternative: https://github.com/mojohaus/versions-maven-plugin/pull/757

slawekjaranowski:
I'm for modelo - we can create XSD schema and documentation automatically.

slawekjaranowski:
I suppose that this one can be closed, when we choose moelo alternative: #757

ajarmoniuk:
> I suppose that this one can be closed, when we choose moelo alternative: #757

of course.

I'm currently working on this regression saw by @sultan -- #760.

The root cause is the fact that I was only adding "-snapshot" in `incrementSegment` to snapshot versions. This was initially done (before this "infinity calculus") because "-snapshot" is treated as the lowest possible version, lower than the release, which acts as a pseudo "minus infinity".

Well, I could either do that or introduce a real "minus infinity".

ajarmoniuk:
The "-segment" approach seems to be fixing the issue; trying the more ambitious approach ;)

Issue #367, opened 2019-09-11
Ignored usage of properties in child POMs


heruan:
I have all version properties defined in the parent POM, then I have several child POMs defining dependency and plugin management.

When running `display-property-updates` on the parent POM, I would expect the plugin to resolve the property usage on its children; instead, it returns:

```
This project does not have any properties associated with versions.
```

Example POMs:

```xml
<project>
    <groupId>my.group</groupId>
    <artifactId>my-parent</artifactId>
    <version>1.0.0</version>
    <packaging>pom</packaging>

    <properties>
        <version.foo>1.0.0</version.foo>
    </properties>
</project>
```

```xml
<project>
    <parent>
        <groupId>my.group</groupId>
        <artifactId>my-parent</artifactId>
        <version>1.0.0</version>
    </parent>
    <artifactId>my-artifact</artifactId>
    <packaging>pom</packaging>

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>foo.group</groupId>
                <artifactId>foo-artifact</artifactId>
                <version>${version.foo}</version>
            </dependency>
        </dependencies>
    </dependencyManagement>
</project>
```

Output of plugin execution:

```
~/my-parent$ mvn versions:display-property-updates
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO]
[INFO] my-parent                                                          [pom]
[INFO] my-artifact                                                        [pom]
[INFO]
[INFO] -------------------------< my.group:my-parent >-------------------------
[INFO] Building my-parent 1.0.0                                           [1/2]
[INFO] --------------------------------[ pom ]---------------------------------
[INFO]
[INFO] --- versions-maven-plugin:2.7:display-property-updates (default-cli) @ my-parent ---
[INFO]
[INFO] This project does not have any properties associated with versions.
[INFO]
[INFO]
[INFO] -----------------------< my.group:my-artifact >-------------------------
[INFO] Building my-artifact 1.0.0                                         [2/2]
[INFO] --------------------------------[ pom ]---------------------------------
[INFO]
[INFO] --- versions-maven-plugin:2.7:display-property-updates (default-cli) @ my-artifact ---
[INFO]
[INFO] This project does not have any properties associated with versions.
[INFO]
[INFO] my-parent .......................................... SUCCESS [  0.484 s]
[INFO] my-artifact ........................................ SUCCESS [  0.007 s]
[...]

PR #759, opened 2022-10-15
#367: Include parent projects in property resolution

ajarmoniuk:
Includes parent projects in property resolution for property updates. Instead of only considering the model of the current project, will also consult parent models to see if there are property definitions for properties defined in the current project.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/api/PomHelper.java
+++ b/src/main/java/org/codehaus/mojo/versions/api/PomHelper.java
@@ -22,11 +22,13 @@ package org.codehaus.mojo.versions.api;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.events.XMLEvent;
 
+import java.io.BufferedReader;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


ajarmoniuk:
@slawekjaranowski Please review.

src/main/java/org/codehaus/mojo/versions/api/PomHelper.java
```diff
@@ -103,15 +105,12 @@ public static Model getRawModel( MavenProject project )
      * @throws IOException if the file is not found or if the file does not parse.
      */
     public static Model getRawModel( File moduleProjectFile )
-        throws IOException
+            throws IOException
     {
-        try ( FileInputStream input = new FileInputStream( moduleProjectFile ) )
+        try ( Reader reader = new BufferedReader( new InputStreamReader( Files.newInputStream( moduleProjectFile
```
slawekjaranowski, line 110: There is `Files.newBufferedReader`

ajarmoniuk:
As you can see, tried that, but it caused the `it-encoding-001` test to fail. `Files.newBufferedReader` assumes `UTF-8` for input, which is not always the case.

madorb:
well this was an unwanted surprise! is there any way to disable this functionality? 

consider users of say, spring boot that have [spring-boot-dependencies](https://search.maven.org/artifact/org.springframework.boot/spring-boot-dependencies/2.7.1/pom) as their parent

it's _quite_ unhelpful to now see the full list of all properties from the parent by default. Most of those are for dependencies any given application doesn't use. 

This radical of a change really should come with a configuration option either in the xml or command line.

ajarmoniuk:
@madorb yes, but only in the upcoming release, see #795

PR #765, opened 2022-10-17
Minor enhancement: use streams API in getNewestUpdates and getVersions

ajarmoniuk:
Uses Streams API to streamline some of the methods, more concise and more readable in my (subjective) opinion.

Does not need to be added to the upcoming release.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
+++ b/src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
@@ -20,11 +20,7 @@ package org.codehaus.mojo.versions.api;
  */
 
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Optional;
-import java.util.Set;
-import java.util.TreeSet;
-import java.util.stream.Collectors;
 
 import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
@@ -35,6 +31,7 @@ import org.codehaus.mojo.versions.ordering.BoundArtifactVersion;
 import org.codehaus.mojo.versions.ordering.InvalidSegmentException;
 import org.codehaus.mojo.versions.ordering.VersionComparator;
 
+import static java.util.Collections.reverseOrder;
 import static java.util.Optional.empty;
 import static java.util.Optional.of;
 import static org.codehaus.mojo.versions.api.Segment.SUBINCREMENTAL;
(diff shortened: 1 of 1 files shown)


src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
```diff
@@ -284,25 +269,13 @@ public final ArtifactVersion[] getVersions( Restriction restriction, boolean inc
...
-            }
-            result.add( candidate );
-        }
-        return result.toArray( new ArtifactVersion[0] );
+        return Arrays.stream( getVersions( includeSnapshots ) )
+                .filter( candidate ->
+                        ( versionRange == null || ArtifactVersions.isVersionInRange( candidate, versionRange ) )
+                        && ( restriction == null || isVersionInRestriction( restriction, candidate ) )
+                        && ( includeSnapshots || !ArtifactUtils.isSnapshot( candidate.toString() ) ) )
+                .collect( () -> new TreeSet<>( getVersionComparator() ), TreeSet::add, Set::addAll )
```
slawekjaranowski, line 277: Can be worked, something like:

```
.sorted( versionComparator )
.distinct()
.collect( toList() )
```

it be more clear what we want to achieve

src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
```diff
@@ -284,25 +269,13 @@ public final ArtifactVersion[] getVersions( Restriction restriction, boolean inc
...
-                continue;
-            }
-            result.add( candidate );
-        }
-        return result.toArray( new ArtifactVersion[0] );
+        return Arrays.stream( getVersions( includeSnapshots ) )
+                .filter( candidate ->
+                        ( versionRange == null || ArtifactVersions.isVersionInRange( candidate, versionRange ) )
+                        && ( restriction == null || isVersionInRestriction( restriction, candidate ) )
+                        && ( includeSnapshots || !ArtifactUtils.isSnapshot( candidate.toString() ) ) )
```
slawekjaranowski, line 276: can we have multiple `filter`, like

```
.filter( candidate -> versionRange == null || ArtifactVersions.isVersionInRange( candidate, versionRange ) )
.filter( candidate -> restriction == null || isVersionInRestriction( restriction, candidate ) )
.filter( candidate -> includeSnapshots || !ArtifactUtils.isSnapshot( candidate.toString() ) ) )
```

What do you think?

ajarmoniuk: Sure it can

src/main/java/org/codehaus/mojo/versions/api/AbstractVersionDetails.java
ajarmoniuk: I thought TreeSet has better performance, but it doesn't really matter in this case 😄

sultan:
I wonder, Instead of sending null ranges, could we send noop filters. And have filters in final fields to avoid multiple instances. What do you think is best ?

ajarmoniuk:
> I wonder, Instead of sending null ranges, could we send noop filters. And functions to create them like filterForXxx. The goal would be to reduce the number of functions maybe. What do you think is best ?

@slawekjaranowski do you think we could have Discussions enabled here? 😄

PR #766, opened 2022-10-17
Build Modello improvement, change namespace for xsd

slawekjaranowski:
- all Modello source will have the same version of model
- generate Modello documentation in model directory - we have the same report name as documentation for model
- change name space to https://www.mojohaus.org/VERSIONS/...
- preserver old version of rule.xsd


src/main/mdo/dependency-updates-report.mdo
```diff
@@ -21,8 +21,8 @@ under the License.
 
 <model xmlns="http://codehaus-plexus.github.io/MODELLO/2.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://codehaus-plexus.github.io/MODELLO/2.0.0 https://codehaus-plexus.github.io/modello/xsd/modello-2.0.0.xsd"
-       xsd.namespace="http://mojo.codehaus.org/versions-maven-plugin/dependency-updates-report/1.0.0"
-       xsd.targetNamespace="http://mojo.codehaus.org/versions-maven-plugin/dependency-updates-report/1.0.0">
+       xml.namespace="https://www.mojohaus.org/VERSIONS/DEPENDENCY-UPDATES-REPORT/${version}"
```
ajarmoniuk, line 24: Do we really want to change the namespace of the reports with every new release of the plugin?

slawekjaranowski: It is `version` of model form `modello-maven-plugin` configuration  - it is not project version

PR #771, opened 2022-10-19
Correcting BoundArtifact comparison and eliminating incrementSegment

ajarmoniuk:
So, I examined BoundArtifactVersion and especially the comparison and tried to rule out/handle all corner cases I could imagine. 

The premise for that was that we were still using "-snapshot" to construct the lower bound such that all other versions would naturally (using how the current implementation of the Maven comparison works) be compared as more major. I didn't think this was good enough and implemented the same using BoundArtifact.

So, my idea was to, instead using "-SNAPSHOT" to denote the lowest possible version on a given segment, use a BoundArtifact representing an upper bound on the less major segment.

So, instead of 1.0.0-SNAPSHOT to denote the least possible version of "1.0.0", we could use something like 0.9.9-(infinity) to achieve the same (so, "attack" from the other side).

Because we're not actually using the `incrementSegment` method anymore, I removed that method. The ideal goal would be to make the "Comparator" classes actually implement the interface only, and not do auxiliary things like count segments. But that was not my goal here.

This obviously does not need to go into the upcoming release. I've tested it, but maybe we should test it a bit more.


src/main/java/org/codehaus/mojo/versions/ordering/BoundArtifactVersion.java
```diff
@@ -74,20 +108,43 @@ protected BoundComparableVersion( BoundArtifactVersion artifactVersion )
...
         {
+            // all segments more or equally major than artifactVersion.segment can change
             return compareTo( ( (List<Item>) items ).iterator(),
                     ( (Iterable<Item>) o.items ).iterator(), artifactVersion.segment.value() );
 
         }
 
-        @SuppressWarnings( "checkstyle:InnerAssignment" )
-        private int compareTo( Iterator<Item> left, Iterator<Item> right, int comparisonLimit )
+        private int compareTo( Iterator<Item> left, Iterator<Item> right, int comparisonsLeft )
```
ajarmoniuk, line 117: This is the most important part of this PR.

ajarmoniuk:
In that case I would consider it for the upcoming release. @slawekjaranowski what do you think?

sultan:
+1! I would go with release your code instead my "increment forces snapshot"

PR #772, opened 2022-10-20
Quality: Refactoring Reports classes

sultan:
the goal would be to shorten render methods by factorizing duplicate code

i'm not fully happy about the end result code so it's in draft but open to feedback/reviews

i believe the current code still gives a good idea of (some) cleaner methods


sultan:
rebased + javadoc for the added methods

sultan:
might close this one without merging if the reports are all translated to mdo.

slawekjaranowski:
Looks ok, thanks.

Issue #730, opened 2022-10-02
New Feature : Exclude intermediate Minors


sultan:
The suggested filter would filter out to exclude verbose intermediate versions, to keep only, in this example :

* 10.0.26
* 10.1.0

the filter/algo would act beginning at the very last, taking minor prefix "10.1." and filter out every other version string starting with "10.1." then finding another Minor, extracting "10.0." and filter out again, etc, until end of list.

![image](https://user-images.githubusercontent.com/5782559/193446173-60c8387b-7b31-4e69-a790-fea03ba22102.png)

sultan:
Result with new code

![image](https://user-images.githubusercontent.com/5782559/193449297-e818fb06-1e64-4e2c-9437-fdbd722a3db0.png)

sultan:
the filter has been rethinked as so:
at most two versions per Major.Minor: latest release + following latest pre-release versions

sultan:
this could be de/activated with a flag
-Dversions.reportLength=FULL
-Dversions.reportLength=SHORT

@ajarmoniuk is there any rull for the flags names?

ajarmoniuk:
Not that I know of :)

ajarmoniuk:
@sultan are you still going to work on this?

sultan:
Still on my todo list. I am having difficulties knowing where to put my filter. I can't get it to be streamable. And not sure where to put it after the api used streams. Any advice appreciated.

PR #773, opened 2022-10-20
Report Verbosity / Concise Reports option 2 - do not merge yet

sultan:
Option 2 for #730 :

Before : 

image

After with like beaks : 

image

After with list separator : 

image

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
+++ b/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
@@ -143,12 +143,6 @@ public abstract class AbstractVersionsReport<T>
     @Parameter( property = "allowSnapshots", defaultValue = "false" )
     protected boolean allowSnapshots;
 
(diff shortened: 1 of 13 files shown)


ajarmoniuk:
I think I like the option with line breaks more.

sultan:
I'd like to have more returns from maintainers and decide aboute the new xml options names. If free on such choice, I might suggest one or two and continue this PR without waiting

ajarmoniuk:
> I'd like to have more returns from maintainers and decide aboute the new xml options names. If free on such choice, I might suggest one or two and continue this PR without waiting

Hi, do you mean to change xml schemas of the reports (so, element and attribute names etc)? Or just options? Especially with the latter – even if you decide to retain the old names for a time being and mark as deprecated, there will immediately be complaints and bug reports about breaking backwards compatibility as soon as you finally remove the old ones. So I personally wouldn't do that. Just my 2 cents.

PR #777, opened 2022-10-21
Resolves #776: onlyUpgradable change the filter to versions where the current version is not the latest one

### LINKED ISSUE #776: onlyUpgradable does not work for dependency-updates-report
Hi.
I use: mvn versions:dependency-updates-report -DonlyUpgradable=true

But the dependency-updates-report.html still contains dependencies with a green checkbox indicating that this dependency is still on the latest version.

Is this a bug or am I missing something?

ajarmoniuk:
Root cause of the issue: `VersionsHelper.lookupDependenciesUpdates` actually returns all dependency versions, and not just updates.

The problem manifested itself when the currently used version was the latest, but there were more versions reported (older than the current one).

This is now corrected, i.e. the method returns only updates, in contrast to `lookupArtifactVersions`.

`onlyUpgradableOnly` looks to work fine now. Updated the unit tests to test also for cases when the version used by an artifact is the latest version, but there are more available versions (older than the current one).

Removed some redundant classes implementing `Callable`, which could very easily be inlined which results in a much cleaner code.

Also, as a minor pedantic fix, corrected the "splitted" [sic] to "split" in one method name.

### DIFF
--- a/src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReportMojo.java
+++ b/src/main/java/org/codehaus/mojo/versions/DependencyUpdatesReportMojo.java
@@ -25,15 +25,18 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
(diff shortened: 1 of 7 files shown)


ajarmoniuk:
@slawekjaranowski Please review and merge.

ajarmoniuk:
Converting to draft -- I'll try to do it properly.

Issue #758, opened 2022-10-14
Modularise the plugin


ajarmoniuk:
Another (mainteinance) idea: modularise the plugin: at least a separate module for the API, etc.

slawekjaranowski:
Idea is known ... topic not easy

https://groups.google.com/g/mojohaus-dev/c/hRltMfTmB4o

We also have some of proposition in PR #619

slawekjaranowski:
The most important is that we will break a links to documentation site.

Site is referenced from many place ... we need some redirection from old url to new one.

There is many useful hints https://gist.github.com/domenic/1f286d415559b56d725bee51a62c24a7

ajarmoniuk:
I don't get why you want to move the repository though. I'll create a draft PR so that you can have a look at my work so far, albeit not done yet.

ajarmoniuk:
> The most important is that we will break a links to documentation site.

I've updated the links, but now I'm actually thinking of maybe changing the `outputDirectory` of `maven-plugin-plugin:report` so that all mojo documentation is generated in the same folder?

slawekjaranowski:
# Proposition for first step

## first modules

 - parent - with gav `org.codehaus.mojo.versions:versions`
 - versions-maven-plugin - `org.codehaus.mojo:versions-maven-plugin` - almost everything as is
 - model - `org.codehaus.mojo.versions:model` - model for main plugin - rules
 - model-report - `org.codehaus.mojo.versions:model-report` - model for reporting
 
### assumptions 
- plugin will has original groupId `org.codehaus.mojo`
- all new modules will have new groupId `org.codehaus.mojo.versions` - we will have everything  for `versions` project under separate groupId in one place
- current documentations is for plugin - so should be in versions-m-p module as is
 
## rename git repository to versions
 
Links like https://www.mojohaus.org/versions-maven-plugin/... should be redirect to https://www.mojohaus.org/versions/versions-maven-plugin/...

I hope it can be achieved by creating simple html files in https://github.com/mojohaus/mojohaus.github.io with redirects 
AFS analytics can help to detect missing items

https://analytics.apache.org/index.php?module=CoreHome&action=index&date=yesterday&period=day&idSite=18#?idSite=18&period=day&date=yesterday&category=General_Actions&subcategory=Events_Events

# Next steps

## Introduce new module `versions-report-maven-plugin`

Reporting needs many stuff on dependencies for doxia and so on - those are not needed for standard plugin

[...]

PR #782, opened 2022-10-23
#758: Implementing the first stage of module split

ajarmoniuk:


ajarmoniuk:
I think we need some kind of a commit freeze in order to commit this -- too many changes.

slawekjaranowski:
I have worked a little on site output.

Please check by:
```
mvn site site:stage
```

and investigate output in `target/staging/versions`

ajarmoniuk:
> I have worked a little on site output.
> 
> Please check by:
> 
> ```
> mvn site site:stage
> ```
> 
> and investigate output in `target/staging/versions`

Maven site won't build anymore.

slawekjaranowski:
> By the way, why did you remove the XSD schemas of the reports from the XSD section? I can re-add those unless you explicitly wanted those removed...

I moved it to place where is generated ... for old versions of XSD we will need place a copy in root site to preserver old path.
We will change repo name ... 😄

slawekjaranowski:
> Looks like maven site requires the artifacts to be present in the local repo. So we need to do install -P run-its instead of verify in maven-verify.yml?

I see - we use under site build current version of plugin ... it is tricky and sometime works, but with more dependencies for another module can stop working ...

To be clear we should used previous version of plugin for site build.

ajarmoniuk:
> > Looks like maven site requires the artifacts to be present in the local repo. So we need to do install -P run-its instead of verify in maven-verify.yml?
> 
> I see - we use under site build current version of plugin ... it is tricky and sometime works, but with more dependencies for another module can stop working ...
> 
> To be clear we should used previous version of plugin for site build

But that won't work - the previous version doesn't have modules 😏

Instead, why don't we do something like an extra Github Ci job before running site - Install with no tests, no enforcer, no chesktyle etc. Or compose that as an extra goal in the Site job?

slawekjaranowski:
I changed to instal for first step build ... I'm waiting for build ...

slawekjaranowski:
I'm want to go with this PR ... even more I can update site with snapshot version before release, so we can quickly resolve all site issues.

Next PR like enforcer rule can be put in separate module - What do you think?

ajarmoniuk:
> I'm want to go with this PR ... even more I can update site with snapshot version before release, so we can quickly resolve all site issues.
> 
> Next PR like enforcer rule can be put in separate module - What do you think?

Definitely!

ajarmoniuk:
Can we merge?

slawekjaranowski:
So, merged ... next release will be very interesting 😄

slawekjaranowski:
eeeh I forgot to squash 😠

Issue #794, opened 2022-10-25
Set goal with newVersion and oldVersion arguments specified fails to update only the versions which matches the oldVersion and updates all the versions


soselamet:
Set goal with newVersion and oldVersion arguments specified fails to update only the versions which matches the oldVersion and updates all the versions.

Example :
command line / mvn versions:set -DnewVersion="TEST" -DoldVersion="TEST2"

Sample maven execution output log lines : 

[INFO] Processing change of com.some.obscure.company.someniftymodule-parent:TEST2 -> TEST
.
.
.
[INFO]     Updating project com.some.obscure.company.someniftymodule-document-parent
[INFO]         from version 0.101.1-SNAPSHOT to TEST

Possible Solution :
Please see the diff file in the zip archive below for the possible fix.

[oldVersion_fix.zip](https://github.com/mojohaus/versions-maven-plugin/files/9862054/oldVersion_fix.zip)

ajarmoniuk:
Would you like to make a PR with your changes?

ajarmoniuk:
I'm unable to reproduce the issue you're reporting. Could you attach a sample project to reproduce it? Also, could you please state the version of the plugin you're working with?

ajarmoniuk:
I was able to reproduce the issue.

It does not occur if the artifact with the given oldVersion is found in the reactor.

Strangely, if the artifact with the given version is *not* found, then the root model is changed anyway! I don't understand the reason for it, need to see *why* this has been done like that. Obviously, as mentioned earlier, your change breaks a few tests. It could have also been done earlier in the stack, in the main method (execute).

However, we're dealing with an old code, and, as I said, I need to undestand the reason it's been done like that first 😆

soselamet:
Thank you for taking this issue in your focus.

The diff I shared was a quick and experimental snippet that was not intended to be a PR level contribution.

I guess integration tests are not executed as part of the regular maven install profile so I didn't observe any test failure. 😃

Please let me know if there is a need for direct contribution via a PR.

PR #799, opened 2022-10-26
#794: SetMojo would always change the version of the POM, regardless if a match was found

ajarmoniuk:
Probably an old technical debt of sorts in SetMojo: the mojo would normally change the files matching oldVersion, groupId, or ArtifactId.

However, if no match was found in the reactor, then the root model is changed anyway!

https://github.com/mojohaus/versions-maven-plugin/blob/b75a5656113f8974e7bb26f2cd0cdeacff59800e/src/main/java/org/codehaus/mojo/versions/SetMojo.java#L384

I don't understand the reason for it. When checked git blame, the original author added those lines to fix some breaking integration tests. However, it looks like no tests fail even with those lines removed.

Anyhow. I don't think that behaviour was intuitive and it is certainly very unexpected.

A side effect of that behaviour was that, with "always", the timestamp of the file was updated regardless if a match was found. So I had to explicitly add all reactor files if the update timestamp switch is set to "always".

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
@@ -33,6 +33,7 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TimeZone;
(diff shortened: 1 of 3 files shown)


ajarmoniuk:
@slawekjaranowski Please review and merge.

src/main/java/org/codehaus/mojo/versions/SetMojo.java
```diff
@@ -375,16 +370,20 @@ public void execute() throws MojoExecutionException, MojoFailureException
...
                             StringUtils.isBlank( oldVersion ) || "*".equals( oldVersion ) ? "" : mVersion );
                 }
             }
-            if ( !found && RegexUtils.getWildcardScore( groupId ) == 0 && RegexUtils.getWildcardScore( artifactId ) == 0
-                && RegexUtils.getWildcardScore( oldVersion ) == 0 )
+
+            if ( "always".equals( updateBuildOutputTimestampPolicy ) )
             {
-                applyChange( project, reactor, files, groupId, artifactId, oldVersion );
+                reactor.values().parallelStream()
```
slawekjaranowski, line 380: What benefit  we have using `parallelStream`?

ajarmoniuk: allows the operation to be performed in parallel (`stream` forces single-threaded execution) and here we don't depend on the order of execution

ajarmoniuk:
Please merge unless you have any more remarks.

ajarmoniuk:
Rebased

PR #804, opened 2022-10-29
Resolves #359: XML Property Updates Report

### LINKED ISSUE #359: XML format for property updates report

ajarmoniuk:
XML version of the Property Updates Report

### DIFF
--- a/versions-maven-plugin/src/it/it-xml-property-updates-report-001/verify.groovy
+++ b/versions-maven-plugin/src/it/it-xml-property-updates-report-001/verify.groovy
@@ -1,0 +1,13 @@
+def pom = new XmlSlurper()
+        .parse( new File( basedir, 'target/property-updates-report.xml' ) )
+
+assert pom.summary.usingLastVersion == 1
+assert pom.summary.nextIncrementalAvailable == 1
+assert pom.summary.nextMinorAvailable == 1
+
+assert pom.properties.property.find { node -> node.propertyName == 'api-version' }
+    .status == 'incremental available'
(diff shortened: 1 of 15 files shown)


ajarmoniuk:
Failing with JVM 8 because of incompatible types: 

java.util.Map cannot be assigned to a java.util.Map argument.

Strange. I've reproduced it with coretto-1.8

sultan:
i also had class cast ex, during report rendering, between classes like Map.Node and Map.TreeNode. i'll open an issue if i can reproduce

versions-maven-plugin/src/it/it-xml-property-updates-report-001/verify.groovy
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return fragment
+    }
+}
+
+def output = new File( basedir, 'target/property-updates-report.xml' ).text
+        .replaceAll( '\\n', '' )
+        .replaceAll( '\\r', '' )
+
+summary = new Fragment( output, '<summary>', '</summary>' )
```
slawekjaranowski, line 29: We can simply use `XmlSlurper`
https://groovy-lang.org/processing-xml.html

PR #805, opened 2022-11-02
Resolves #505: getReactorModels using correct module paths when the module name includes pom.xml

### LINKED ISSUE #505: Plugin cannot update child versions when pom path specified
Project structure:
```
.
└── parent
    ├── moduleA
    │   └── submoduleB
```

parent pom.xml
```
  <artifactId>parent</artifactId>
  <packaging>pom</packaging>
  <version>1.0-SNAPSHOT</version>
  <modules>
    <module>moduleA/pom.xml</module>
  </modules>
```

moduleA pom.xml
```
  <parent>
    <artifactId>parent</artifactId>
    <version>1.0-SNAPSHOT</version>
  </parent>

  <artifactId>moduleA</artifactId>
  <packaging>pom</packaging>
  <modules>
    <module>submoduleB/pom.xml</module>
  </modules>
```

submoduleB pom.xml
```
  <parent>
    <artifactId>moduleA</artifactId>
    <version>1.0-SNAPSHOT</version>
  </parent>

  <groupId>org.example.moduleA</groupId>
  <artifactId>submoduleB</artifactId>
```

Command:
```
mvn org.codehaus.mojo:versions-maven-plugin:2.8.1:set -DnewVersion=123456
```

Expected behaviour:
All `1.0-SNAPSHOT` versions to be replaced with `123456`

What happens:
Only versions in root pom.xml and moduleA/pom.xml is updated, the one in submoduleB/pom.xml stays the same

Note:
When pom.xml paths not specified, plugin works correctly and updates all of them successfully

ajarmoniuk:
Nothing fancy or spectacular, just a simple bug fix for the edge case of having the module name include the pom.xml file path. In the unpatched implementation, the "/pom.xml" was treated as part of the directory name when descending down the module tree to retrieve child models... Fixed that.

Added some tests (unit test for the PomHelper method + integration test for the use case from the issue).

### DIFF
--- a/versions-maven-plugin/src/it/it-set-issue-505/pom.xml
+++ b/versions-maven-plugin/src/it/it-set-issue-505/pom.xml
@@ -1,0 +1,13 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>default-group</groupId>
+    <artifactId>default-artifact</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>moduleA/pom.xml</module>
+    </modules>
+
+</project>
--- a/versions-maven-plugin/src/it/it-set-issue-505/invoker.properties
+++ b/versions-maven-plugin/src/it/it-set-issue-505/invoker.properties
@@ -1,0 +1,2 @@
+invoker.goals = ${project.groupId}:${project.artifactId}:${project.version}:set
+invoker.mavenOpts = -DnewVersion=TEST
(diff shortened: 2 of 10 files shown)


ajarmoniuk:
`getProjectParent( final ModifiedPomXMLEventReader pom, VersionsHelper helper )` and `removeMissingChildModules( Log logger, MavenProject project, Collection<String> childModules )` were no longer used -> removed

ajarmoniuk:
Rebased.

versions-maven-plugin/src/it/it-set-issue-505/pom.xml
```diff
@@ -0,0 +1,13 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>default-group</groupId>
+    <artifactId>default-artifact</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>moduleA/pom.xml</module>
```
slawekjaranowski, line 10: I can have:
```
<module>moduleA/pom-with-special-name.xml</module>
```
what then?

Please add such case in one module.

ajarmoniuk: Good point

rshkv:
I think there might have been a break here: `versions:set` stopped working in my repo and I narrowed it down to this commit. I'm attaching `--debug` logs for a run with this commit and a run on 2.13.0.

[versions-set-output-2.13.0.txt](https://github.com/mojohaus/versions/files/10230777/versions-set-output-good.txt)
[versions-set-output-805.txt](https://github.com/mojohaus/versions/files/10230779/versions-set-output-bad.txt)

slawekjaranowski:
please follow #848 #849

ajarmoniuk:
@rshkv thanks. Tested this with the PR, problem looks solved.

rshkv:
Thank you for the reply guys!

Issue #516, opened 2021-12-13
documentation: it is not obvious what is the difference between release and version


ineffective:
I'm using versions plugin to update versions of libraries in my project. I have single master pom.xml with 17 sub-modules. I asked on [stack-overflow](https://stackoverflow.com/questions/70306046/maven-versions-plugin-horribly-slow) how to speed up artifact resolution. Granted, question sounds aggressive - I'm sorry, it was written in a heat of the moment, because it takes around 40 minutes to check all dependencies. I run two commands (copied from the question):

```shell
mvn -B -U versions:use-releases -DexcludeReactor=false -DgenerateBackupPoms=false -DallowSnapshots=false -Dexcludes='org.springframework.boot:spring-boot-starter-parent:*,log4j:log4j:*,commons-io:commons-io:*,commons-beanutils:commons-beanutils:*'
mvn -B -U versions:use-latest-releases -DexcludeReactor=false -DgenerateBackupPoms=false -DallowSnapshots=false -Dexcludes='org.springframework.boot:spring-boot-starter-parent:*,log4j:log4j:*,commons-io:commons-io:*,commons-beanutils:commons-beanutils:*'
```
It was suggested (by @khmarbaise) that I should use `use-latest-versions` rather than `use-latest-releases`.

The problem: I have no idea and were unable to find on maven's documentation page what is the difference between the two, except that one uses latest versions and the other uses latest release. But what is the difference between release and version? Considering that I use `allowSnapshots=false` there should be (probably) no real difference between the two - and yet there apparently is.

My suggestion is that it should be documented somewhere on plugin's page what these two mean.

ajarmoniuk:
`UseLatestVersions` is an almost exact copy of `UseLatestReleases` code-wise (which will be handled in a separate issue).

There's one difference. `UseLatestReleases` will only process dependencies matching the pattern:

`^(.+)-((SNAPSHOT)|(\d{8}\.\d{6}-\d+))$`

so, snapshots or versions denoted with a *yyyy-MM-dd* suffix.

I'll add that to the documentation.

ajarmoniuk:
To be done together with #807

ineffective:
So in simple words "version" is "anything" and release is "not-a-snapshot"? Thanks, that makes it MUCH clearer.

PR #808, opened 2022-11-04
Resolves #807: Abstracting a common base for UseLatestVersionsMojo and UseLatestReleasesMojo

### LINKED ISSUE #807: Maintenance: code duplication between UseLatestVersions and UseLatestReleases
`UseLatestVersions` is an almost exact copy of `UseLatestReleases` code-wise. Possibly there are more examples of such duplicated classes, to be checked.

ajarmoniuk:
- Abstracting a common base for UseLatestVersionsMojo and UseLatestReleasesMojo.
- Elaborating on the difference between a "release" and a "version"

Two issues:
#807 
#516


ajarmoniuk:

ajarmoniuk:
My own afterthought after this is merged is that, as with other mojos. Across the whole plugin we basically have a host of different features, which may be turned on and off. I think I'll need to review the whole of the plugin and use a pattern like the decorator so that we can simply compose the behaviour of the mojo of different behaviours, depending on which features are enabled. See #289

PR #813, opened 2022-11-09
Resolves #289: Handle processParent in mojos which support it.

ajarmoniuk:
Also:
- added failIfNotReplaced to ForceReleasesMojo as a first steps towards converging with UseReleasesMojo
- added unit tests and integration tests
- some refactoring

More steps towards folding ForceReleases and UseReleases will follow. Also, I see it's possible to reuse parts of the common base class used for lastest versions/releases, but this will follow in future PRs.


ajarmoniuk:

ajarmoniuk:
I have updated it a bit to fold `UseReleasesMojo` with `ForceReleasesMojo` a bit further, but that's about it in this PR, I'll continue this subject in a separate PR.

Another point to consider: I think #134 is not limited to that one mojo, but may be present everywhere -- to be checked and possibly fixed.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/reporting/AbstractVersionsReportRenderer.java
```diff
@@ -64,10 +64,6 @@
     protected ArtifactVersionsCache allUpdatesCache
             = new ArtifactVersionsCache( AbstractVersionDetails::getAllUpdates );
 
-    /**
```
ajarmoniuk, line 67: Removed this documentation comment as it was misleading (not used by Plexus) and not really necessary -- the parameters are pretty obvious.

ajarmoniuk:
Sorry about these multiple updates today, I think that's it for this PR.

Now it all looks lean and clean. Almost all mojos updating versions consist of an almost similar method (which by the way could also be folded), and a (small) business method, providing the provider method, instructing `useLatestVersions` how exactly it should retrieve the latest version.

ajarmoniuk:
@slawekjaranowski rebased. Please review.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractVersionsUpdaterMojo.java
```diff
@@ -205,10 +205,11 @@
     @Parameter( property = "changeRecorderOutputFile",
                 defaultValue = "${project.build.directory}/versions-changes.xml" )
     private File changeRecorderOutputFile;
+
     /**
      * The change recorder implementation.
      */
-    private ChangeRecorder changeRecorder;
+    private volatile ChangeRecorder changeRecorder;
```
slawekjaranowski, line 212: Is it accessed from multi thread somehow?

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractVersionsUpdaterMojo.java
```diff
@@ -560,24 +574,21 @@ else if ( PomHelper.setPropertyVersion( pom, version.getProfileId(), property.ge
...
+    protected ChangeRecorder getChangeRecorder()
     {
         if ( changeRecorder == null )
         {
-            if ( "none".equals( this.changeRecorderFormat ) )
-            {
-                changeRecorder = ChangeRecorderNull.create();
-            }
-            else if ( "xml".equals( this.changeRecorderFormat ) )
+            synchronized ( this )
```
slawekjaranowski, line 582: why is needed ... streams..parallel ?

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractVersionsUpdaterMojo.java
ajarmoniuk: Not needed indeed. Removed.

slawekjaranowski:
As I assume - after it PR #289 an issue #288 - can be closed - right?

Issue #743, opened 2022-10-07
`use-latest-releases` checks transitive dependencies instead of direct dependencies


AdrienHorgnies:
Given the following `pom.xml` file : 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter-parent</artifactId>
        <version>2.7.4</version>
        <relativePath/> <!-- lookup parent from repository -->
    </parent>
    <groupId>com.example</groupId>
    <artifactId>demo</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <name>demo</name>
    <description>demo</description>
    <properties>
        <java.version>17</java.version>
    </properties>
    <dependencies>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter</artifactId>
        </dependency>

        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-test</artifactId>
            <scope>test</scope>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
            </plugin>
        </plugins>
    </build>

</project>
```

I expected the command :

    mvn org.codehaus.mojo:versions-maven-plugin:2.12.0:use-latest-releases -DallowMajorUpdates=false -DgenerateBackupPoms=false

to give the following input :

    [INFO] --- versions-maven-plugin:2.12.0:use-latest-releases (default-cli) @ demo ---
    [INFO] Minor version changes allowed
    [INFO] artifact org.springframework.boot:spring-boot-starter-parent: checking for updates from central

But instead it checks all the following dependencies : 

    [INFO] --- versions-maven-plugin:2.12.0:use-latest-releases (default-cli) @ demo ---
    [INFO] Minor version changes allowed
    [INFO] artifact io.netty:netty-transport-sctp: checking for updates from central
[...]

PR #476, opened 2021-08-02
Handle import scope dependencies in dependencyManagement (multiple goals) (Fixes multiple issues #308, #309, #346, #395)

frode-carlsen:
Fixes #308, Fixes #309, Fixes #346, Fixes #395 for following goals
* use-dep-version
* use-latest-versions
* use-latest-snapshots
* use-latest-release
* use-next-releases
* use-next-versions
* use-next-snapshots

Issue #395, opened 2020-05-11
use-dep-version (and other goals) does not handle the import-scoped dependency


JohnWu-Pro:
* `versions:use-dep-version` does not update the version of the import-scoped dependency, in the `<dependencyManagement>` section. Other goals may also not handle import-scoped dependency correctly.
* This issue might be related to [Include import-scoped dependency in dependency plugin goals](https://issues.apache.org/jira/browse/MDEP-694)

janhicken:
This seems to be a duplicate of #346. 
The `PomHelper` already features a method to find the imported POMs inside the dependency management section. This is just not used in many of this plugin's goals. The commit referenced in #346 shows a possible fix for the goal `use-latest-releases`.

cemo:
Same for `versions:compare-dependencies. I would like to compare my dependencies which is imported into the project with the remote pom.

frode-carlsen:
I've createad a PR that also will update the import scoped dependencies for the following goals:
* use-dep-version
* use-latest-versions
* use-latest-snapshots
* use-latest-release
* use-next-releases
* use-next-versions
* use-next-snapshots

Issue #346, opened 2019-04-21
use-latest-releases ignores import scoped BOM dependencies


asw56:
use-latest-releases ignores import scoped BOM dependencies

Apache Maven 3.6.0
versions-maven-plugin 2.7

example:

```
     <dependencyManagement>
         <dependencies>
             <dependency>
                 <groupId>org.springframework</groupId>
                 <artifactId>spring-framework-bom</artifactId>
                 <version>4.0.1.RELEASE</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
         </dependencies>
     </dependencyManagement>
 
     <dependencies>
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-context</artifactId>
         </dependency>
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-web</artifactId>
         </dependency>
     </dependencies>
```

Expected result: spring-framework-bom managed version to be updated (5.1.6.RELEASE at time of writing)
Actual result: no change to pom.xml

frode-carlsen:
The ```PomHelper.readImportedPOMsFromDependencyManagementSection```  only handles imported poms, not regular ones. I've created an updated PR here https://github.com/mojohaus/versions-maven-plugin/pull/476

PR #814, opened 2022-11-11
Issue 134 removing the no longer needed patch + adding integration tests

ajarmoniuk:
Looks like the patch is no longer needed. Tested with maven 3.2.5 as well as 3.6.3. Added integration tests which test against the case described in the issue.

Some explanation.

The removed method extracted the imported dependencies from the raw model, which were then used *in addition to* dependencies from the imported boms, which were added by Maven.

But, because the dependencies added by Maven are not present in the POM to be updated, processing them is simply wasted resources. Instead, it's much better to just process the raw model.


ajarmoniuk:
Waiting for #476

ajarmoniuk:
Looks like infrastructure problems again.

Could you please relaunch the failed jobs?

ajarmoniuk:

ajarmoniuk:
Also resolves #743

ajarmoniuk:
Resolves: #743, #308, #309, #346, #395

versions-maven-plugin/src/it/it-unlock-snapshots-issue-134/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+  <version>1.0</version>
+  <packaging>pom</packaging>
+  <name>Check if boms are upated</name>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.codehaus.plexus</groupId>
+        <artifactId>plexus-utils</artifactId>
+        <version>1.5.1-20090311.051742-1</version>
```
slawekjaranowski, line 16: Even if today such version of snapshot exists - can be removed, why not use some of dumy dependencies?

versions-maven-plugin/src/it/it-use-dep-version-issue-134/pom.xml
```diff
@@ -0,0 +1,24 @@
...
+  <artifactId>it-use-releases-issue-134</artifactId>
+  <version>1.0</version>
+  <packaging>pom</packaging>
+  <name>Check if boms are upated</name>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.maven</groupId>
+        <artifactId>maven-parent</artifactId>
```
slawekjaranowski, line 15: why not use some of dumy dependencies?

versions-maven-plugin/src/it/it-unlock-snapshots-issue-134/pom.xml
ajarmoniuk: I coundt reproduce it with dummy dependencies. I did use dummy dependencies wherever I could.

ajarmoniuk: As an alternative or a temporary workaround, I can remove these tree integration tests using non-dummy artifacts. Should I?

versions-maven-plugin/src/it/it-use-dep-version-issue-134/pom.xml
ajarmoniuk: Fixed

versions-maven-plugin/src/it/it-unlock-snapshots-issue-134/pom.xml
ajarmoniuk: Fixed

ajarmoniuk:

slawekjaranowski:
> Waiting for #476

rather replace - not waiting - I'm right?

ajarmoniuk:
> > Waiting for #476
> 
> rather replace - not waiting - I'm right?

Not anymore. I meant to wait for the outcome of the discussion there.

PR #817, opened 2022-11-19
Make includeParent false by default to retain backward compatibility.

ajarmoniuk:
There have been compelling remarks that includeParent should be false, otherwise we're breaking backwards compatibility. I agree. Hence this small PR.


ajarmoniuk:
@slawekjaranowski @slachiewicz please review

pafnucy-tech:
I just needed to downgrade to 2.13 as the plugin looked for the modules in the pom directory instead of one level higher. Is this related?

PR #819, opened 2022-11-20
Introduce Plugin API for ChangeRecorder

slawekjaranowski:
TODO
 - manual - how to write plugin extensions
 - document new code


versions-api/src/main/java/org/codehaus/mojo/versions/api/recording/ChangeRecord.java
```diff
@@ -0,0 +1,40 @@
...
+        UPDATE_PARENT( "updateParent" ),
+        UPDATE_PROPERTY( "updateProperty" ),
+        USE_DEPENDENCY_VERSION( "useDependencyVersion" ),
+        USE_LATEST_RELEASES( "useLatestReleases" ),
+        USE_LATEST_SNAPSHOTS( "useLatestSnapshots" ),
+        USE_LATEST_VERSIONS( "useLatestVersions" ),
+        USE_NEXT_RELEASES( "useNextReleases" ),
+        USE_NEXT_SNAPSHOTS( "useNextSnapshots" ),
+        USE_NEXT_VERSIONS( "useNextVersions" ),
+        USE_RELEASES( "useReleases" );
```
slawekjaranowski, line 22: There is a list of label used as `kind` - for me it is more source of change ... maybe not needed at all

slawekjaranowski:
I see that `ChangeRecorder` is only used when change is introduced to project.
But we can see that `changeRecorder*` parameters are also available for display mojos, so users can be confused.

slawekjaranowski:
@ajarmoniuk - Pleas look.

versions-api/src/main/java/org/codehaus/mojo/versions/api/recording/ChangeRecord.java
ajarmoniuk: What if we used this like the logger name in SLF4J and similar logger interfaces? Maybe make it a free-form text and use the class name with it, just like for loggers? What do you think?

versions-common/src/main/java/org/codehaus/mojo/versions/recording/ChangeRecorderXML.java
```diff
@@ -72,31 +82,40 @@ public ChangeRecorderXML()
...
-                                    final String oldVersion, final String newVersion )
+    public final void recordUpdate( ChangeRecord changeRecord )
     {
-        Objects.requireNonNull( kind, "kind" );
-        Objects.requireNonNull( groupId, "groupId" );
-        Objects.requireNonNull( artifactId, "artifactId" );
-        Objects.requireNonNull( oldVersion, "oldVersion" );
-        Objects.requireNonNull( newVersion, "newVersion" );
-
         final Element update = this.document.createElementNS( CHANGES_NAMESPACE, "update" );
```
ajarmoniuk, line 87: How about using Modello? 😄

ajarmoniuk (commented):
Nice. How about using Modello for building the xml file? 😄

versions-common/src/main/java/org/codehaus/mojo/versions/recording/ChangeRecorderXML.java
slawekjaranowski: I saw it.
Yes Modello should be used here for consistencies - now this report is undocumented.
Out of scope for this change.

ajarmoniuk: Fair enough.

versions-api/src/main/java/org/codehaus/mojo/versions/api/recording/ChangeRecord.java
slawekjaranowski: So if we use class name it will be rather then source of change than kind ...

slawekjaranowski: Maybe `kind` should be where version is changed: dependency, dependencyMangmen, plugin, pluginManagmen, properties, parent

ajarmoniuk: Good idea. +1

slawekjaranowski: done

ajarmoniuk (approved):
Looks good. Can we also have a link from the main index.md page (versions-maven-plugin/src/site/index.md) to the new module?

PR #828, opened 2022-11-24
#704: Removing ArtifactResolver

ajarmoniuk:
Second stage: Removing ArtifactResolver.

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
@@ -47,14 +47,10 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
-import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.artifact.manager.WagonManager;
-import org.apache.maven.artifact.repository.ArtifactRepository;
-import org.apache.maven.artifact.resolver.ArtifactNotFoundException;
 import org.apache.maven.artifact.resolver.ArtifactResolutionException;
-import org.apache.maven.artifact.resolver.ArtifactResolver;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.execution.MavenSession;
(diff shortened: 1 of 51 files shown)


ajarmoniuk:
Also managed to reduce the `settings` property (it's already available via `MavenSession`).

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -479,13 +464,21 @@ private List<IgnoreVersion> getIgnoredVersions( Artifact artifact )
...
-                        : mavenSession.getCurrentProject().getRemoteArtifactRepositories(),
-                localRepository );
+        try
+        {
+            aetherRepositorySystem.resolveArtifact( mavenSession.getRepositorySession(),
+                    new ArtifactRequest( toArtifact( artifact ),
+                            usePluginRepositories
+                                    ? mavenSession.getCurrentProject().getRemotePluginRepositories()
+                                    : mavenSession.getCurrentProject().getRemoteProjectRepositories(),
+                            getClass().getName() ) );
```
slawekjaranowski, line 476: We need process response and update source artifact, like:
```
rtifactResult artifactResult  = resolveArtifact( ... );

        artifact.setFile( artifactResult.getArtifact().getFile() );
        artifact.setVersion( artifactResult.getArtifact().getVersion() );
        artifact.setResolved( true );
```

slawekjaranowski (changes requested):

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
ajarmoniuk: Done

ajarmoniuk:
Changes applied.

ajarmoniuk:
@slawekjaranowski rebased and resolved conflicts

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -479,13 +465,30 @@ private List<IgnoreVersion> getIgnoredVersions( Artifact artifact )
...
+            artifact.setFile( artifactResult.getArtifact().getFile() );
+            artifact.setVersion( artifactResult.getArtifact().getVersion() );
+            artifact.setResolved( artifactResult.isResolved() );
+            if ( !artifactResult.getExceptions().isEmpty() )
+            {
+                getLog().warn( "Exceptions thrown while trying to resolve artifact " + artifact );
+                artifactResult.getExceptions().forEach( e ->
+                        getLog().warn( "\t" + e.getMessage() ) );
+            }
```
slawekjaranowski, line 486: I think that never occurs in this place. 

When exceptions list is set also exception is thrown.

ajarmoniuk: It looks like you're right. Removed.

slawekjaranowski (approved):

PR #831, opened 2022-11-26
#704: Remove WagonManager

ajarmoniuk:
Removing legacy `WagonManager` and going forward with `Transporter` from https://maven.apache.org/resolver/index.html

Advantage: multiple transporter providers possible, not just the default `Wagon`. Also, future-proof.

`it-transporter-002` is testing FileTransporter instead of the default Wagon.


ajarmoniuk:

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -868,19 +686,172 @@ else if ( !excludePropertiesList.isEmpty() && excludePropertiesList.contains( pr
...
+                        catch ( Exception e )
+                        {
+                            log.warn( "Error while reading the rules string: " + e.getMessage() );
+                            return null;
+                        }
+                    } )
+                    .filter( Objects::nonNull )
+                    .findFirst()
+                    .orElse( null );
+        }
```
slawekjaranowski, line 845: I'm not sure if we can use Resolver Transports in such way.
Rather it is for internal use by Resolver.

In this case - we simply need download a resource from remote url.
Better will be use eg HttpClient from HttpComponents.

ajarmoniuk: Hmm, are you sure? If that is the case, then I'll provably need to use different transports depending on protocol (file, classpath, etc).

ajarmoniuk: Hmm, are you sure? If that is the case, then I'll provably need to use different transports depending on protocol (file, classpath, etc).

slawekjaranowski: @cstamas - can have a best knowledge about it.

ajarmoniuk: If not, HttpClient5 is lightweight enough and I'd like to use the Fluent flavour of it, if you don't mind.

ajarmoniuk: Posted a question to dev@maven.apache.org, but I can't see it. Do I need an invite?

slawekjaranowski: HttpClient5 will be ok.

I would like to introduce next util class for getting resource from remote, with method like:

```
 ... downolad( uri, serverId, mavenSession ) 
```

in method we can find proper server configuration from `mavenSession.settings`
Maven settings can also has a proxy configuration which should be under consideration

ajarmoniuk: I was thinking of something like that actually. E.g. provide several implementations per protocol (classpath, file, http/https) and use a factory method to select the matching implementation from a map.

Or use Wagon without WagonManager instead? Since Wagon basically already provides such a tnin layer, currently using HttpClient 4.5. Is Wagon going to be retired altogether?

ajarmoniuk: @hboutemy Do you know if we can continue to use Wagon or is it going to be retired and we should be using e.g. HttpClient directly?

slawekjaranowski: > Posted a question to [dev@maven.apache.org](mailto:dev@maven.apache.org), but I can't see it. Do I need an invite?

You should first subscribe to list.

slawekjaranowski: ? should works .... after send email to: dev-subscribe at maven.apache.org

ajarmoniuk:
Still waiting for an answer whether we can simply use Wagon (without WagonManager from maven-compat). Note that there's also a org.apache.maven.repository.legacy.WagonManager, of which I don't know whether it's going to be deprecated or not. Hence my question about Wagon in general.

For the time being, I'm still working on a version using HttpClient (largely based on Wagon), but I think I'll also create a version using Wagon only while waiting for the answer.

ajarmoniuk:
Answer from the current Wagon maintainer is that it is definitely future-proof. 
Got an implementation using Wagon ready.

However, I'm still thinking whether not to simply employ HttpClient 5 instead. Disadvantage of such a solution would be that it would basically boil down to be a yet another reinterpretation of Wagon... 

What do you think, @slawekjaranowski @slachiewicz ?

ajarmoniuk:
Rebased and resolved conflicts (a lot of them again). Still weighing in whether I should go with this or simply HttpClient5.

ajarmoniuk:
As you can see -- implementation using Wagon is ready and kicking, but I am still thinking whether not to go with pure HttpClient.

ajarmoniuk:
I've decided that it's not worth it to go with HttpClient5 directly. Too much adapting plus Wagon is already a thin layer around it. Maybe later.

if you decide it's ready, please merge

slawekjaranowski:
Ok, good enough

ajarmoniuk:
Still working on a version with HttpClient5. But that will come later. Perhaps in the next release.

PR #832, opened 2022-11-26
Clarification in README - maintained versions, issues and PR creation

slawekjaranowski:
According to clarification we can only test with the latest Maven on GitHub it will save time for executing

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -25,9 +25,7 @@ jobs:
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v3
     with:
       ff-goal: '-P run-its install'         # site use project version for reporting
-      ff-maven: "3.8.6"                     # Maven version for fail-fast-build
       verify-goal: '-P run-its install'     # should be the same as for first build
-      maven-matrix: '[ "3.2.5", "3.8.6" ]'  # Maven versions matrix for verify builds
 
   deploy:
     name: Deploy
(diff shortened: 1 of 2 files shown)


slawekjaranowski:
What do you think @slachiewicz, @stefanseifert, @ajarmoniuk, @sultan

sultan:
i would say let's move forward and be clear on the minimal maven version needed to run the plugin.
any chance we move to maven 4 when available?

ajarmoniuk:
I think we need to check the minimal version supported as well as a current version - which can be the latest release or some other one we choose. But without testing with the minimal version we can not be sure if we really support it.

slachiewicz:
Maybe we should explicitly say that we recommend Maven 3.8.6 (plus link) and add note to verify if issue can be reproduced also with latest version of Maven

slawekjaranowski:
> i would say let's move forward and be clear on the minimal maven version needed to run the plugin. any chance we move to maven 4 when available?

Maven 4 API is not compatible backward to Maven 3, so it is for discussion later.

slawekjaranowski:
> I think we need to check the minimal version supported as well as a current version - which can be the latest release or some other one we choose. But without testing with the minimal version we can not be sure if we really support it.

I know that when we will testing only with latest versions of Maven some of issue can be detected. 

But only checking minimum version doesn't make us sure that will be work with all other older versions like 3.3.x, 3.5.x, 3.6.x

When someone detect an issue with Maven 3.5.x - I want not spend time for resolving it.

slawekjaranowski:
> Maybe we should explicitly say that we recommend Maven 3.8.6 (plus link) and add note to verify if issue can be reproduced also with latest version of Maven

Exactly it is my idea and my proposition in change.

PR #833, opened 2022-11-27
#704: Remove remaining ArtifactRepository leftovers + use version range for VersionsHelper where applicable

ajarmoniuk:
Utilising the fact that we can pass the *versionRange* directly to `RepositorySystem.resolveVersionRange`, which should optimise the search process (possibly less lookups performed).

Also, removing the unused `findLatestVersion( Artifact, VersionRange,                                              Boolean, boolean )` method.


ajarmoniuk:
I've noticed that there are still some `ArtifactRepository` leftovers left. Removed those.

Also, Maven 3 version range resolver allows version range resolving -- using that to speed up version resolution a bit.

Small refactoring/cleanup, especially in DisplayPluginUpdates.

@slawekjaranowski please check

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -194,19 +204,19 @@ public DisplayPluginUpdatesMojo( RepositorySystem repositorySystem,
      * @throws MojoExecutionException when things go wrong.
      */
     private Map<String, String> getSuperPomPluginManagement()
-        throws MojoExecutionException
+            throws MojoExecutionException
     {
-        if ( new DefaultArtifactVersion( "3.0" ).compareTo( runtimeInformation.getApplicationVersion() ) <= 0 )
+        if ( runtimeInformation.isMavenVersion( "[3,)" ) )
```
slawekjaranowski, line 209: It always be true ... we require 3.2.5 as minimum

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -194,19 +204,19 @@ public DisplayPluginUpdatesMojo( RepositorySystem repositorySystem,
...
         {
             getLog().debug( "Using Maven 3.x strategy to determine superpom defined plugins" );
             try
             {
                 Method getPluginsBoundByDefaultToAllLifecycles =
-                    LifecycleExecutor.class.getMethod( "getPluginsBoundByDefaultToAllLifecycles",
-                            String.class );
+                        LifecycleExecutor.class.getMethod( "getPluginsBoundByDefaultToAllLifecycles",
+                                String.class );
```
slawekjaranowski, line 216: Such reflections is probably not needed

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -408,11 +418,11 @@ public void execute()
...
 
         Set<Plugin> plugins = getProjectPlugins( superPomPluginManagement, parentPlugins, parentBuildPlugins,
-                                                 parentReportPlugins, pluginsWithVersionsSpecified );
+                parentReportPlugins, pluginsWithVersionsSpecified );
 
         List<String> pluginUpdates = new ArrayList<>();
         List<String> pluginLockdowns = new ArrayList<>();
-        ArtifactVersion curMavenVersion = runtimeInformation.getApplicationVersion();
+        ArtifactVersion curMavenVersion = new DefaultArtifactVersion( runtimeInformation.getMavenVersion() );
         ArtifactVersion specMavenVersion = MinimalMavenBuildVersionFinder.find( getProject(), "2.0", getLog() );
```
slawekjaranowski, line 426: To investigate if code is needed for Maven 3.x

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -1111,76 +1069,58 @@ private Map<String, Plugin> getLifecyclePlugins( MavenProject project )
...
+     * @throws MojoFailureException thrown if plugin retrieval did not succeed
      */
     // pilfered this from enforcer-rules
     // TODO coordinate with Brian Fox to remove the duplicate code
     private Set<Plugin> getBoundPlugins( MavenProject project, String thePhases )
-        throws PluginNotFoundException, LifecycleExecutionException, IllegalAccessException
+            throws MojoExecutionException
     {
-        if ( new DefaultArtifactVersion( "3.0" ).compareTo( runtimeInformation.getApplicationVersion() ) <= 0 )
+        if ( runtimeInformation.isMavenVersion( "[3,)" ) )
```
slawekjaranowski, line 1079: We use 3.2.5

ajarmoniuk:
As advised in https://cwiki.apache.org/confluence/display/MAVEN/Plugin+migration+to+Maven3+dependencies

> Remove [maven-compat](https://maven.apache.org/ref/current/maven-compat/apidocs/) (or give it the test-scope if it is required by the maven-plugin-testing-harness)

ajarmoniuk:
I guess I can move to PlexusJunit4TestCase in another PR.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
ajarmoniuk: Changed from `2.0` to `3.2.5` as that is the minimum version. This is still necessary if the user is running a older than current version of Maven, but > 3.2.5

versions-maven-plugin/src/it/it-display-plugin-updates-001/pom.xml
```diff
@@ -9,7 +9,7 @@
   <name>display-plugin-updates</name>
 
   <prerequisites>
-    <maven>2.0.6</maven>
+    <maven>3.2.5</maven>
   </prerequisites>
```
slawekjaranowski, line 13: prerequisites is only used for maven-plugin packaging.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java
```diff
@@ -442,7 +398,7 @@ public void execute()
             String effectiveVersion = version;
 
             Artifact artifactRange = getHelper().createPluginArtifact( plugin.getGroupId(), plugin.getArtifactId(),
-                                                                       version );
+                    version );
```
slawekjaranowski, line 401: I start thinking to introduce spotless as is done in Maven from parent poms 38 when we have such line changed 😄

slawekjaranowski (approved):

Issue #744, opened 2022-10-07
`display-dependency-updates` version comparison algorithm bug for `-pfd`


garretwilson:
I have the following managed dependency in my Maven POM:

```
<dependency>
  <groupId>javax.faces</groupId>
  <artifactId>javax.faces-api</artifactId>
  <version>2.3</version>
  <scope>provided</scope>
</dependency>
```

However when I invoke `mvn versions:display-dependency-updates`, Versions Maven Plugin shows me:

```
javax.faces:javax.faces-api ........................... 2.3 -> 2.3-pfd
```

Offhand I don't know what `-pfd` means (pre-final-deliverable?), but that's irrelevant. You can see [on Maven Central](https://search.maven.org/artifact/javax.faces/javax.faces-api) that this version was released _before_ v2.3.

The [semantic versioning specification](https://semver.org/) says that _any_ `-*` suffix (which would include `-pfd`) is considered a "pre-release" version. Thus `2.3-pfd` is considered a pre-release version of `2.3` and should not be listed as a new available version. (The specification then goes on to explain how to calculate version precedence for pre-release suffixes, but that's even not the case here, as _any_ pre-release version should not appear greater than the release version.)

(On a related note, the specification also says that version metadata, that is suffixes beginning with `+*`, must be completely ignored when determining version precedence.)

ajarmoniuk:
Confirmed

garretwilson:
Thank you, @ajarmoniuk . It's really so nice to see this project is getting quick and useful responses. I sincerely appreciate it. Have a good weekend.

ajarmoniuk:
No worries! Looking into it. Have a nice weekend too!

ajarmoniuk:
Apparently, the problem also exists in Maven itself -- looks like the class `ComparableVersion` got carried over to `org.apache.maven.artifact.versioning.ComparableVersion` there.

If you use a range specification like

```xml
<dependency>
  <groupId>javax.faces</groupId>
  <artifactId>javax.faces-api</artifactId>
  <version>[2.3,)</version>
  <scope>provided</scope>
</dependency>
```

your project will actually depend on version 2.3-pfd (you can check with e.g. `dependency:tree`). But this is indeed incorrect.

ajarmoniuk:
https://issues.apache.org/jira/browse/MNG-7559

mprins:
[...]

PR #835, opened 2022-11-30
#744 fix maven version comparison

sultan:
see:
* PR #744

### DIFF
--- a/versions-maven-plugin/src/test/java/org/codehaus/mojo/versions/UpdateParentMojoTest.java
+++ b/versions-maven-plugin/src/test/java/org/codehaus/mojo/versions/UpdateParentMojoTest.java
@@ -204,6 +204,7 @@ public class UpdateParentMojoTest {
         assertThat(newVersion, nullValue());
     }
 
+    /*
     @Test
     public void testAllowSnapshots()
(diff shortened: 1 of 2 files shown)


versions-maven-plugin/src/test/java/org/codehaus/mojo/versions/UpdateParentMojoTest.java
```diff
@@ -204,6 +204,7 @@ public void testParentDowngradeForbiddenWithRange()
         assertThat(newVersion, nullValue());
     }
 
+    /*
```
slawekjaranowski, line 207: Why disable this test?

In such case I prefer to delete code instead of commented out. 

When we want to left commented code we should add information why is commented, what is plan to restore, etc ...

slawekjaranowski (changes requested):
Can you prepare a unit test for `ComparableVersion`.

It looks like we can do a test with examples how specific version should be compared - maybe parameterized.

ajarmoniuk:
Hmm, just a thought. Keep in mind that a change in the versioning algorithm will possibly cause bug reports with people complaining about the changed behaviour.

sultan:
there is no need for this PR anymore as the mojo plugin uses ComparableVersion from Maven now

Issue #793, opened 2022-10-24
add an option to exclude parent from properties updates


jonathanlermitage:
Since version 2.13.0, the `display-property-updates` goal shows many properties updates for Spring Boot based projects.  
Spring Boot projects use a Spring Boot parent pom which declares a BOM with many dependencies and their corresponding version properties.  
Even if we don't actually use these dependencies (they are declared in a `dependencyManagement` block), the `display-property-updates` goals now shows updates for all of these properties.  
I think it would be useful to provide a parameter that allows us to ignore version properties updates from parent's pom. If this parameter already exists, could you please update the documentation https://www.mojohaus.org/versions-maven-plugin/display-property-updates-mojo.html?  
Thanks!

How to reproduce: create a sample Maven project with https://start.spring.io/. You can also test this project https://github.com/jonathanlermitage/spring-mvc-tiny-sample. Then, invoke `display-property-updates` and see the proposed properties updates.  
With plugin 2.12.0 I get:  
```
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------------< coin.coin:spring-mvc-tiny-sample >------------------
[INFO] Building spring-mvc-tiny-sample 0.0.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- versions-maven-plugin:2.12.0:display-property-updates (default-cli) @ spring-mvc-tiny-sample ---
[INFO] Major version changes allowed
[INFO] Major version changes allowed
[INFO] Major version changes allowed
[INFO] 
[INFO] The following version property updates are available:
[INFO]   ${guava.version} ............................... 28.1-jre -> 31.1-jre
[INFO]   ${rest-assured.version} .............................. 4.1.1 -> 5.2.0
[INFO]   ${versions-maven-plugin.version} ................... 2.12.0 -> 2.13.0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.963 s
[INFO] Finished at: 2022-10-24T11:08:17-04:00
[INFO] ------------------------------------------------------------------------
```

With plugin 2.13.0 I get:  
```
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------------< coin.coin:spring-mvc-tiny-sample >------------------
[INFO] Building spring-mvc-tiny-sample 0.0.1-SNAPSHOT
[...]

PR #838, opened 2022-12-01
Resolves #837: Add includeParent (default true) to UpdatePropertiesMojo

### LINKED ISSUE #837: UpdatePropertiesMojo processes parent POMs
UpdatePropertiesMojo traverses parent POMs in order to find and update properties. However, this is not necessary, only adds to chatter of the mojo, since parent POMs won't be able to be updated anyway.

As a side thing, change recorder does contain changes even if the POM has not been modified.

Seen in #793

ajarmoniuk:
Upon request of a user in #793, adding `includeParent` (`true` by default) to `UpdatePropertiesMojo`.

Setting it to `false` will speed up execution to a great extend without sacrificing accuracy in most cases. However, if the POM redefines properties used in parent POMs, dependencies bound to those properties will not be discovered, so the plugin will not redefine those properties. Therefore the default value is `true`.

### DIFF
--- a/versions-maven-plugin/src/it/it-update-properties-issue-837/invoker.properties
+++ b/versions-maven-plugin/src/it/it-update-properties-issue-837/invoker.properties
@@ -1,0 +1,2 @@
+invoker.goals = ${project.groupId}:${project.artifactId}:${project.version}:update-properties
+invoker.mavenOpts = -DincludeParent=false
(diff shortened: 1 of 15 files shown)


ajarmoniuk:

This resolves #838

ajarmoniuk:
Added the same to `UpdatePropertyMojo`

ajarmoniuk:
And abstracted a common base for the two mojos.

ajarmoniuk:
Tried to resolve conflicts using GitHub 🤦

slawekjaranowski:
We set such parameters to `false` in #817

Should be here default as `true`?

ajarmoniuk:
> Should be here default as `true`?

No. With `false`, it will break #582. Like I said in the first comment,

> Setting it to false will speed up execution to a great extend without sacrificing accuracy in most cases. However, if the POM redefines properties used in parent POMs, dependencies bound to those properties will not be discovered, so the plugin will not redefine those properties. Therefore the default value is true.

So, it can only be treated as a speedup if one knows what they're doing. It will sometimes render the results wrong.

slawekjaranowski:
Maven 2.x code should be remove, can be done in separate PR.
We use 3.2.5 as minimum.

ajarmoniuk:
> Maven 2.x code should be remove, can be done in separate PR.
We use 3.2.5 as minimum.

I'll do that in another PR then

Issue #333, opened 2019-02-19
aggregate reports support


davewichers:
Can the maven versions plugin be enhanced so users can generate aggregate reports for all the sub-modules in a multimodule project? Similar to: https://maven.apache.org/plugins/maven-checkstyle-plugin/checkstyle-aggregate-mojo.html

This users wants it to, but no answer to his query: https://stackoverflow.com/questions/49171182/maven-multi-module-project-versionsdependency-updates-report 

If there is no way to easily generate a single aggregate report, can there be an option to at least interlink the reports, or generate a root .html report that links them all together?

thomasvamos:
I would aprecciate this feature as well.

davewichers:
Why was this auto closed? Seems like a useful feature.

davewichers:
It's been THREE years! Any chance this could get implemented?

ajarmoniuk:
Good idea.

Question: should this also be supported for Display...Updates mojos? Strangely enough -- descending into the reactor tree is being done e.g. in the Set mojo which could result in inactive profiles being processed. 

@slachiewicz @slawekjaranowski opinions?

PR #841, opened 2022-12-05
#333: aggregated reports site information

MihaStemberger:
Adding aggregate plugins information to site


slawekjaranowski:
@MihaStemberger thanks

PR #842, opened 2022-12-06
Wagon should use remote repository instead of proxy repository for repo authentication

ajarmoniuk:
Quick fix, fixing a typo. No unit tests since it would require lots of mocking, and it was not possible to inject the RemoteRepositorySession object. And a WireMock test would be too much of an overkill for this IMO.


ajarmoniuk:

PR #845, opened 2022-12-11
Enable Spotless plugin - automatic code formatting

slawekjaranowski:


ajarmoniuk:
Does it mean maven 2022 code style as well?

slawekjaranowski:
> Does it mean maven 2022 code style as well?
> 
> Edit: looks like it 👍

Exactly

PR #856, opened 2022-12-15
Resolves #855: Set should evaluate expressions

### LINKED ISSUE #855: ${revision} not being updated
This is a sample pom to illustrate the issue
`
    <project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

    <artifactId>test-versions</artifactId>
    <version>${revision}</version>

    <properties>
        <maven.compiler.source>17</maven.compiler.source>
        <maven.compiler.target>17</maven.compiler.target>
        <revision>1.3</revision>
    </properties>

    <build>
        <pluginManagement>
            <plugins>
                <plugin>
                    <version>2.14.1</version>
                </plugin>
            </plugins>
        </pluginManagement>
    </build>
</project>`

`mvn versions:set -DnewVersion=2.0`

With plugin 2.13.0 it resolves all ${revision} to the new version.
With plugin 2.14.0 and 2.14.1 it does nothing.

ajarmoniuk:
Fixing a regression after #799 

The premise of that PR was that the mojo was updating every file if it **did not** find any match anywhere. This made the mojo update files where e.g. the version was specified as a parameter. 

The fix does it differently -- by resolving parameters and checking if they match the artifact coordinates.

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
@@ -39,6 +39,7 @@ import java.util.SortedMap;
 import java.util.TimeZone;
 import java.util.TreeMap;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.model.Model;
@@ -65,6 +66,7 @@ import org.codehaus.plexus.components.interactivity.Prompter;
 import org.codehaus.plexus.components.interactivity.PrompterException;
 import org.codehaus.plexus.util.StringUtils;
 
+import static java.util.Optional.ofNullable;
 import static org.codehaus.plexus.util.StringUtils.isEmpty;
 
 /**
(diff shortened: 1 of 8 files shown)


ajarmoniuk:

ajarmoniuk:
@pzygielo @slachiewicz @slawekjaranowski do you think there might be other corner cases to check here?

pzygielo:
@ajarmoniuk thanks for asking, but I'm not advanced user of versions-m-p nor I know the codebase well.
It looks that feedback from community is pretty fast though :smile:

ajarmoniuk:
Sławek's question is valid though -- perhaps this PR is not needed after all.
On the other hand, it does fix the issue :)

slawekjaranowski:
a Corner cases from https://maven.apache.org/maven-ci-friendly.html

Property as:
```
 <properties>
    <revision>1.0.0-${buildNumber}-SNAPSHOT</revision>
  </properties>
```

version with multiple properties:
```
<version>${revision}${sha1}${changelist}</version>
```

Multi Module example

ajarmoniuk:
@slawekjaranowski added these corner cases

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
```diff
@@ -348,9 +350,22 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                     Pattern.compile(RegexUtils.convertWildcardsToRegex(fixNullOrEmpty(oldVersion, "*"), true));
 
             for (Model m : reactor.values()) {
-                final String mGroupId = PomHelper.getGroupId(m);
-                final String mArtifactId = PomHelper.getArtifactId(m);
-                final String mVersion = PomHelper.getVersion(m);
+                Map<String, String> properties = ofNullable(m.getProperties())
+                        .map(p -> p.entrySet().parallelStream()
+                                .collect(Collectors.toMap(e -> e.getKey().toString(), e -> e.getValue()
+                                        .toString())))
```
slawekjaranowski, line 356: From: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-

> The returned Collector is not concurrent. For parallel stream pipelines, the combiner function operates by merging the keys from one map into another, which can be an expensive operation. If it is not required that results are inserted into the Map in encounter order, using [toConcurrentMap(Function, Function)](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toConcurrentMap-java.util.function.Function-java.util.function.Function-) may offer better parallel performance.

So for such operation where not many elements are to compute we can use simple stream instead of parallel Stream.

I hope that we will have a less problems with simple stream.

ajarmoniuk: Done.

slawekjaranowski:
On more test will be appreciable:

Parent with property:

```
 <parent>
    <groupId>org.apache.maven.ci</groupId>
    <artifactId>ci-parent</artifactId>
    <version>${revision}</version>
  </parent>
```

ajarmoniuk:
Assuming you meant a parent with a property with a child module -- added.

Issue #858, opened 2022-12-16
Versions Maven Plugin updates random Module subsets with 2.14.1


jwachter:
Environment
-------------

- Maven Multimodule Project with 295 Modules in a strict parent-child hierarchy
- Maven Wrapper using Maven 3.8.6
- Calling `mvnw.cmd` or `./mvnw` with `versions:set -DnewVersion=1.2.3-SNAPSHOT -DgenerateBackupPoms=false`

Observed Behavior
-------------------

On multiple runs it is not deterministic which versions are updated:

- Run 1: 135 of 295 are not updated after BUILD SUCCESS (All children of a sub-parent)
- Run 2: 0 of 295 are not updated after BUILD SUCCESS
- Run 3: 3 of 295 are not updated after BUILD SUCCESS
- Run 4: 138 of 295 are not updated after BUILD SUCCESS (Children in multiple disjoint sub-parent hierarchies)

One of the sub parent hierarchies that seems to be always involved (partially) when something is not updated contains a single level of ~140 children but still not always only that hierarchy or a specific number of the children is affected.

With 2.14.0 it seems to not be broken yet for the same amount of test runs.

As this seems somewhat like a typical race condition we suspect that the introduction of  `parallelStream` in this line combined with a non-concurrent Map might be problematic: https://github.com/mojohaus/versions/commit/c74ced27d704d6a3d1bad9c916da3203e170f2de#diff-f58ee99d4e29219ee1519ba59712674919933915a6485ab987093718d0b773dbR1359

Expected Behavior
-------------------

All modules are consistently updated.

ajarmoniuk:
Thanks for the bug report. I think it's rather `(files::add)` which causes the problem in combination with `parallelStream`.

ajarmoniuk:
@corebonts there's also https://github.com/mojohaus/versions/blob/master/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java#L375 

here, the non-thread safe collection is modified from within a parallel stream

please add that to your branch if you want

PR #859, opened 2022-12-16
#858 Fix child module resolution for multi-level projects

corebonts:
Child modules were processed in a parallel stream which added elements to an unsynchonized LinkedHashMap.

This may resulted an incomplete set of child modules.

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/PomHelper.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/PomHelper.java
@@ -1356,7 +1356,7 @@ public class PomHelper {
 
         File baseDir = model.getPomFile().getParentFile();
 
-        getAllChildModules(model, logger).parallelStream()
+        getAllChildModules(model, logger).stream()
                 .map(moduleName -> new File(baseDir, moduleName))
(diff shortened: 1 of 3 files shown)


ajarmoniuk:
@corebonts please also modify https://github.com/mojohaus/versions/blob/master/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java#L375

ajarmoniuk:
Closed my own PR fixing this. @slachiewicz @slawekjaranowski could you please approve the run? Thanks

slachiewicz (approved):

slawekjaranowski:
I also see that `parallelStream` is used in many other places 

I don't see a much performance benefit of used it but can cause a problem like here.
Maybe we simply can replace all occurrences.

ajarmoniuk:
Went other these during the evaluation of this one. Basically a problem occurs when there is a modification of a non-thread safe collection during stream execution. There should be no problem with the creation of a new collection, which is the case in all other places than the corrected in this PR.

corebonts:
After Andrzej told me about the other problematic class, I also went through all the other usages and I can confirm that the others are ok for now. But I neither see a big performance benefit of using parallel streams.

PR #861, opened 2022-12-17
Removes parallelStream use with non-thread safe collectors

ajarmoniuk:
Relates to #855, reviewed and removed unsafe parallelStream usage where the terminator was not thread safe.


ajarmoniuk:
@slawekjaranowski

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -194,7 +194,7 @@ public ArtifactVersions lookupArtifactVersions(
                                                             .getRemoteProjectRepositories(),
                                             "lookupArtifactVersions"))
                             .getVersions()
-                            .parallelStream()
+                            .stream()
```
ajarmoniuk, line 197: here it was not strictly necessary since noneMatch should work

PR #863, opened 2022-12-18
Resolves #474: Using raw model for dependencies in UseDepVersion

### LINKED ISSUE #474: Add support for property defined version update in the goal use-dep-version
I noticed that currently `use-dep-version` works only when the version is defined inline, as in the following works:
```
mvn versions:use-dep-version  -DdepVersion=2.3.1 '-Dincludes=javax.xml.bind:jaxb-api'
```

```
--- a/example/pom.xml
+++ b/example/pom.xml
@@ -80,7 +80,7 @@
         <dependency>
             <groupId>javax.xml.bind</groupId>
             <artifactId>jaxb-api</artifactId>
-            <version>2.3.0</version>
+            <version>2.3.1</version>
         </dependency>
```

But the following property defined version doesn't get updated:
```
    <properties>
      <jaxb.version>2.3.0</jaxb.version>
    </properties>

        <dependency>
            <groupId>javax.xml.bind</groupId>
            <artifactId>jaxb-api</artifactId>
            <version>${jaxb.version}</version>
        </dependency>
```

This is a feature request to make `use-dep-version` work for property defined dependencies too.

Use case: we have many internal repositories which depends on few internal common libraries, sometimes we would want to update all of those repos to the latest version of the common libraries for bugs/security issues etc. But the downstream maven projects sometimes would have the version defined as a property or sometimes as inline version. If one command could handle both cases we could easily update all the repos to the target version.

ajarmoniuk:
UseDepVersion mojo did not allow changing the dependency for dependencies which used a property.

For a reason: this basically does not work with the interpolated model: *PomHelper.setDependencyVersion* works by comparing the text of the POM file with the dependency coordinates and will replace the dependency when a string comparison match is found. 

If we work on the interpolated model, the properties will be resolved, thus the dependency version will also be set and will no longer use a property. `UseDepVersion` **will** call *PomHelper.setDependencyVersion*, but it will not be able to retrieve the dependency to be changed, since the POM file still contains the unresolved property.

To mitigate that, here we are using the unresolved, raw model to do that. It does the job.

What's more, I think that the same problem exists everywhere else. I'm not changing it everywhere else, because I think we need to give it a thought. 

Anwyay, please review and merge @slawekjaranowski

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseDepVersionMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseDepVersionMojo.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.DependencyManagement;
+import org.apache.maven.model.Model;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Mojo;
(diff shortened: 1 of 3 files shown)


ajarmoniuk:
Converting to draft -- I think this is still incorrect. If a dependency is defined via properties, the property value should be updated, and not the dependency.

ajarmoniuk:
Changing the underlying property version makes the scope much broader: the whole project tree needs to be scanned since a property can be defined in the parent, or e.g. redefined in a child.

ajarmoniuk:
Still working on this. I'll reopen the PR when ready.

PR #867, opened 2022-12-19
Fixes #866: Require maven 3.2.5

### LINKED ISSUE #866: The plugin should print an excplicit error message if the maven version is lower than the 3.2.5
There's too many confused users reporting guice errors raised because the new components cannot be resolved. I think returning an error message could clear the confusion a bit.

ajarmoniuk:
@slawekjaranowski obvious


pom.xml
```diff
@@ -94,6 +94,10 @@
     </contributor>
   </contributors>
 
+  <prerequisites>
+    <maven>3.2.5</maven>
```
pzygielo, line 98: This has to be in the plugin, not the top-level pom.

Issue #869, opened 2022-12-20
Performance issue with many versions/artifacts


TobiX:
Creating update reports with many potential update versions & artefacts is pretty slow. As an example take this `pom.xml`:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>test</groupId>
  <artifactId>test</artifactId>
  <version>1-SNAPSHOT</version>
  <packaging>pom</packaging>
  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
    <awssdk.version>2.13.26</awssdk.version>
    <!-- awssdk.version>2.17.290</awssdk.version -->
  </properties>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>software.amazon.awssdk</groupId>
        <artifactId>bom</artifactId>
        <version>${awssdk.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
  <build>
    <pluginManagement>
      <plugins>
        <plugin>
          <artifactId>maven-site-plugin</artifactId>
          <version>3.12.1</version>
        </plugin>
        <plugin>
          <groupId>org.codehaus.mojo</groupId>
          <artifactId>versions-maven-plugin</artifactId>
          <version>2.14.1</version>
          <configuration>
            <formats>
              <format>html</format>
              <format>xml</format>
            </formats>
          </configuration>
        </plugin>
      </plugins>
    </pluginManagement>
  </build>
  <reporting>
    <excludeDefaults>true</excludeDefaults>
    <plugins>
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>versions-maven-plugin</artifactId>
      </plugin>
    </plugins>
  </reporting>
</project>
```

This uses a fairly outdated version of the [AWS SDK bom](https://central.sonatype.dev/artifact/software.amazon.awssdk/bom/2.19.0/versions). This bom contains a lot of modules, which all individually are checked for newer versions. This takes ages.

[...]

PR #870, opened 2022-12-20
Add a simple cache for ComparableVersions

TobiX:
This is a simple fix for #869.

This can improve performance in pathetic corner cases by up to 50%. From some example executions (something similar as the pathetic case in #869):

Before:
```
[INFO] Total time:  01:51 min
```

After:
```
[INFO] Total time:  43.714 s
```

In this first implementation, the cache is unbounded. In all my tests it never grew over ~1500 entries, but it might keep some memory active in pathetic cases...

PS: I used a ConcurrentHashMap, because I don't know how parallel (if at all?) Maven can run this code...

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/ordering/ComparableVersion.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/ordering/ComparableVersion.java
@@ -26,8 +26,10 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Properties;
 import java.util.Stack;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Generic implementation of version comparison.
@@ -37,6 +39,8 @@ import java.util.Stack;
  * Note: The implementation of the maven core should be used.
  */
 public class ComparableVersion implements Comparable<ComparableVersion> {
+    private static final Map<String, ComparableVersion> CACHE = new ConcurrentHashMap<>();
+
     private String value;
 
     private String canonical;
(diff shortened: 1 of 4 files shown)


versions-common/src/main/java/org/codehaus/mojo/versions/ordering/ComparableVersion.java
```diff
@@ -283,6 +287,19 @@ public String toString() {
...
+        return CACHE.computeIfAbsent(version, ComparableVersion::new);
+    }
+
+    /**
+     * Create a ComparableVersion from a string.
+     *
+     * @deprecated Use {@link #of(String)} instead to use an internal cache.
+     */
+    @Deprecated
     public ComparableVersion(String version) {
```
slawekjaranowski, line 303: Is there this constructor usage somewhere?
If no I would to consider to make a private it instead deprecating.

TobiX: BoundComparableVersion extends ComparableVersion, so we could make the constructor protected?

slawekjaranowski: > BoundComparableVersion extends ComparableVersion, so we could make the constructor protected?

ok

TobiX: Changed to protected.

ajarmoniuk:
I was looking into moving this to Maven level. However:

Neither ComparableVersion nor DefaultArtifactVersion are immutable: parseVersion will mutate their state. If this is the case, they mustn't be cached. 

In the plugin, we must therefore forbid ComparableVersion users from ever using parseVersion() on the instance retrieved from cache.

TobiX:
@ajarmoniuk Oh right, `parseVersion` is public. That's unfortunate (aka. a bug waiting to happen 😉)... I found no local users of that method, but it might be considered API-breaking to make it private... Would an "immutable" subclass be a viable solution?

ajarmoniuk:
@TobiX well, that's indeed unfortunate. I wanted to promote your cache to maven-core, but looked into it closer and discovered that it's in fact mutable (because of that method). The interface is a bit... botched and it would be really nice to change it, but - Maven is big and who knows who depends on it.

What's more -- I saw that we could actually cache DefaultArtifactVersion instances, since they are also immutable except for that unfortunate method, which is also part of the interface. And we're actually constructing a lot more of those objects, and use them for comparison.

So, instead, I've introuced this change: https://github.com/mojohaus/versions/pull/898

we're caching DefaultArtifactVersions instead and only in the plugin -- on condition that we will never call `parseVersion` in our plugin.

ajarmoniuk:
FYI: https://issues.apache.org/jira/browse/MNG-7659

PR #873, opened 2022-12-22
Resolves #872: Make allowSnapshots an explicit argument in lookupDependencyUpdates and in reports

### LINKED ISSUE #872: Inconsistency between dependency-updates and property-updates
Even if `allowSnapshots` is true, `display-dependency-updates` only detects newer dependencies up to the last release (even if there are newer snapshots), while `display-property-updates` is able to report the newer snapshots

For instance, if the "Snapshots" repository has com.example:foo:2.9.1-SNAPSHOT and 2.9.2-SNAPSHOT and the "Releases" repository has version 2.9.1, I get 
```
mvn versions:display-dependency-updates
[INFO] The following dependencies in Dependencies have newer versions:
[INFO]   com.example:foo .... 2.9.1-SNAPSHOT -> 2.9.1

mvn versions:display-property-updates
[INFO] The following version property updates are available:
[INFO]   ${com-example-foo.version} ......... 2.9.1-SNAPSHOT -> 2.9.2-SNAPSHOT
```

``` 
<repositories>
    <repository>
		<id>Releases</id>
		<url>https://example.com/releases</url>
	</repository>
	<repository>
		<id>Snapshots</id>
		<url>https://example.com/snapshots</url>
		<snapshots>
			<enabled>true</enabled>
		</snapshots>
		<releases>
			<enabled>false</enabled>
		</releases>
	</repository>
</repositories>

...

<dependency>
	<groupId>com.example</groupId>
	<artifactId>foo</artifactId>
	<version>${com-example-foo.version}</version>					
</dependency>

...
		
<plugin>
	<version>2.14.1</version>
	<configuration>
		<allowSnapshots>true</allowSnapshots>
	</configuration>
</plugin>				
```

ajarmoniuk:
allowSnapshots was not an explicit argument in many retrieval operations. Instead, it was assumed that the user would call `AbstractVersionDetails.setIncludeSnapshots()` to set it. This class was used as a container for this value so that it could be passed to report generators.

So, I'm clearing this up and also passing the value **explicitly** where needed.

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
@@ -91,7 +91,6 @@ public class ArtifactVersions extends AbstractVersionDetails implements Comparab
         versionComparator = other.versionComparator;
         versions = other.versions;
         setCurrentVersion(other.getCurrentVersion());
-        setIncludeSnapshots(other.isIncludeSnapshots());
     }
 
     @SuppressWarnings("checkstyle:InnerAssignment")
@@ -122,7 +121,7 @@ public class ArtifactVersions extends AbstractVersionDetails implements Comparab
 
         return new EqualsBuilder()
                 .append(getArtifact(), that.getArtifact())
-                .append(getVersions(), that.getVersions())
+                .append(getVersions(true), that.getVersions(true))
                 .append(getVersionComparator(), that.getVersionComparator())
                 .isEquals();
     }
@@ -131,7 +130,7 @@ public class ArtifactVersions extends AbstractVersionDetails implements Comparab
     public int hashCode() {
         return new HashCodeBuilder(17, 37)
                 .append(getArtifact())
-                .append(getVersions())
+                .append(getVersions(true))
                 .append(getVersionComparator())
                 .toHashCode();
     }
@@ -216,6 +215,24 @@ public class ArtifactVersions extends AbstractVersionDetails implements Comparab
                 : versions.stream()
                         .filter(v -> !ArtifactUtils.isSnapshot(v.toString()))
                         .toArray(ArtifactVersion[]::new);
+    }
+
+    /**
+     * Says whether the versions present in the {@link ArtifactVersions} collection are empty.
+     * If {@code includeSnapshots} is {@code true}, snapshots will not counted, which means that
+     * the method will only count release versions.
+     *
+     * @param includeSnapshots {@code includeSnapshots} is {@code true}, snapshots will not counted, which means that
+     *      * the method will only count release versions.
+     * @return if {@code includeSnapshots} is {@code true}, returns {@code true} if there are no versions.
+     * if {@code includeSnapshots} is {@code false}, returns {@code true} if there are no releases.
+     */
+    public boolean isEmpty(boolean includeSnapshots) {
+        return includeSnapshots
+                ? versions.isEmpty()
+                // the below means: the only versions that could be present in the collection
+                // are snapshots
+                : versions.stream().map(Object::toString).allMatch(ArtifactUtils::isSnapshot);
     }
 
     public VersionComparator getVersionComparator() {
--- a/versions-common/src/test/java/org/codehaus/mojo/versions/api/ArtifactVersionsTest.java
+++ b/versions-common/src/test/java/org/codehaus/mojo/versions/api/ArtifactVersionsTest.java
@@ -49,7 +49,8 @@ public class ArtifactVersionsTest {
 
     @Test
     public void test4DigitVersions() throws Exception {
-        ArtifactVersion[] versions = versions("1.0.0.1", "1.0.0.2", "2.121.2.1", "2.100.0.1", "3.1.0.1", "1.1.1");
+        ArtifactVersion[] versions =
+                versions("1.0.0.1", "1.0.0.2", "2.121.2.1", "2.100.0.1", "3.1.0.1", "1.1.1", "2.0.0-SNAPSHOT");
         final DefaultArtifact artifact = new DefaultArtifact(
                 "group",
                 "artifact",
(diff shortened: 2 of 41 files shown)


ajarmoniuk:

versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
```diff
@@ -122,7 +121,7 @@ public boolean equals(Object o) {
 
         return new EqualsBuilder()
                 .append(getArtifact(), that.getArtifact())
-                .append(getVersions(), that.getVersions())
+                .append(getVersions(true), that.getVersions(true))
```
slawekjaranowski, line 124: default value for field `includeSnapshots` was false - why here is true?

versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
```diff
@@ -131,7 +130,7 @@ public boolean equals(Object o) {
     public int hashCode() {
         return new HashCodeBuilder(17, 37)
                 .append(getArtifact())
-                .append(getVersions())
+                .append(getVersions(true))
```
slawekjaranowski, line 133: Do we change default value?

versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
```diff
@@ -218,6 +217,12 @@ public ArtifactVersion[] getVersions(boolean includeSnapshots) {
                         .toArray(ArtifactVersion[]::new);
     }
 
+    public boolean isEmpty(boolean includeSnapshots) {
+        return includeSnapshots
+                ? versions.isEmpty()
+                : versions.stream().map(Object::toString).allMatch(ArtifactUtils::isSnapshot);
```
slawekjaranowski, line 223: if includeSnapshots is true we check if any versions are present - nevermind snapshot or not 
but if includeSnapshots is false we check if all versions are snapshots ...

maybe some comments for this methods - what does do

versions-common/src/test/java/org/codehaus/mojo/versions/api/ArtifactVersionsTest.java
```diff
@@ -64,21 +64,22 @@ public void test4DigitVersions() throws Exception {
         assertEquals("artifact", instance.getArtifactId());
         assertEquals("group", instance.getGroupId());
         assertThat(
-                instance.getVersions(),
+                instance.getVersions(true),
                 arrayContaining(versions("1.0.0.1", "1.0.0.2", "1.1.1", "2.100.0.1", "2.121.2.1", "3.1.0.1")));
```
slawekjaranowski, line 68: test for `false` value will be appreciated

versions-common/src/test/java/org/codehaus/mojo/versions/api/ArtifactVersionsTest.java
```diff
@@ -64,21 +64,22 @@ public void test4DigitVersions() throws Exception {
...
         assertThat(
-                instance.getVersions(new DefaultArtifactVersion("1.1"), null),
+                instance.getVersions(new DefaultArtifactVersion("1.1"), null, true),
```
slawekjaranowski, line 70: maybe also some snapshot version

versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
ajarmoniuk: ArtifactVersions always contains all versions of the artifact. No matter what the value of "includeSnapshots" was before I removed it. includeSnapshots acted as an additional value to apply on top of the collected versions.

But here, we are checking if two object instances are equal. If we set this to false, an instance with and an instance without snapshots will be equal, incorrectly.

versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
ajarmoniuk: Same as above -- I'm not changing the default value.

versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersions.java
ajarmoniuk: This means -- if include snapshots is false -- then **if there are any versions that are there, all of them are snapshots**

ajarmoniuk: I'll add the following comment:

```
  // the below means: the only versions that could be present in the collection
  // are snapshots
```

ajarmoniuk: Also added javadoc. 

Instead of this, `getVersions().length == 0` was used, which acted the same.

versions-common/src/test/java/org/codehaus/mojo/versions/api/ArtifactVersionsTest.java
ajarmoniuk: Added. Also for isEmpty().

versions-common/src/test/java/org/codehaus/mojo/versions/api/ArtifactVersionsTest.java
ajarmoniuk: Added

versions-enforcer/src/main/java/org/apache/maven/plugins/enforcer/MaxDependencyUpdates.java
```diff
@@ -243,6 +243,13 @@ public class MaxDependencyUpdates implements EnforcerRule2 {
      */
     protected RuleSet ruleSet;
 
+    /**
+     * Whether snapshots should be counted as updates. Default is {@code false}.
+     *
+     * @since 2.15.0
```
slawekjaranowski, line 249: so postpone for next minor release 😄

ajarmoniuk: Should I change this to 2.14.2?

slawekjaranowski: yes, please

ajarmoniuk: Ok, done.

PR #878, opened 2022-12-23
Resolves #474: Add property support to UseDepVersion

### LINKED ISSUE #474: Add support for property defined version update in the goal use-dep-version

ajarmoniuk:
- adding a new parameter, processProperties, which will also update property values if dependencies are controlled by properties
- the above parameter will be turned off in order to retain backwards compatibility

The behaviour of the property is described here:

```
/**
 * <p>Will augment normal processing by, if a dependency value is set using a property, trying to update
 * the value of the property.</p>
 * <p>If the property value is specified directly, will process it normally (as with {@code processProperties} being
 * {@code false}. If the property being updated is redefined in the reactor tree, will only change the property
 * value which lies closest to the dependency being updated. If the same property is also used to set
 * the value of another dependency, will not update that property value, and log a warning instead.
 * Finally, if the property value is specified in a parent file which is outside of the project, will log
 * a warning message.</p>
 * <p>Default is {@code false}.</p>
 *
 * @since 2.14.2
 */
```

also, the behaviour is also "described" in the unit tests.

By the way, I've put 2.14.2 tentatively, but because of the nature of this PR, I'd understand if it did not fit into the release.

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseDepVersionMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseDepVersionMojo.java
@@ -22,27 +22,46 @@ package org.codehaus.mojo.versions;
 import javax.inject.Inject;
 import javax.xml.stream.XMLStreamException;
 
+import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.model.Dependency;
-import org.apache.maven.model.DependencyManagement;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.ModelBase;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.repository.RepositorySystem;
 import org.apache.maven.wagon.Wagon;
-import org.codehaus.mojo.versions.api.ArtifactVersions;
 import org.codehaus.mojo.versions.api.PomHelper;
 import org.codehaus.mojo.versions.api.VersionRetrievalException;
 import org.codehaus.mojo.versions.api.recording.ChangeRecord;
 import org.codehaus.mojo.versions.api.recording.ChangeRecorder;
 import org.codehaus.mojo.versions.rewriting.ModifiedPomXMLEventReader;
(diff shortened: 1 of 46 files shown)


ajarmoniuk:
Manual testing...

ajarmoniuk:
Changed it so that it doesn't scan the whole tree if we're not processing properties.

Unfortunately, there's one problem I still don't know how to tackle -- maven unnecessarily descends into children even if the plugin could have made all necessary changes starting from root level. 

Do you know how to change that so that Maven doesn't run recursively -- from the plugin?

slawekjaranowski:
I would like to postpone this for next minor release ... in next year 😄

ajarmoniuk:
That's OK 😄

ajarmoniuk:
So I guess that's a nice PR for 🎄😄

ajarmoniuk:
Note to self: https://maven.apache.org/developers/mojo-api-specification.html#The_Descriptor_and_Annotations

Since there's time, I'll polish it a bit.

ajarmoniuk:
Changed to aggregate mode; this change spurred the creation of a new enhancement: #880

ajarmoniuk:
Temporarily converting to draft: let's see if I we can use maven (to do the recursion) and plugin context instead.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseDepVersionMojo.java
```diff
@@ -62,12 +81,28 @@ public class UseDepVersionMojo extends AbstractVersionsDependencyUpdaterMojo {
...
+     * the value of the property.</p>
+     * <p>If the property value is specified directly, will process it normally (as with {@code processProperties} being
+     * {@code false}. If the property being updated is redefined in the reactor tree, will only change the property
+     * value which lies closest to the dependency being updated. If the same property is also used to set
+     * the value of another dependency, will not update that property value, and log a warning instead.
+     * Finally, if the property value is specified in a parent file which is outside of the project, will log
+     * a message.</p>
+     * <p>Default is {@code false}.</p>
+     *
+     * @since 2.14.2
```
slawekjaranowski, line 101: I hope next version will be 2.15.0

PR #897, opened 2023-01-02
Fixes #896: Fixing available version retrieval; enabling allowSnapshots

### LINKED ISSUE #896: display-parent-updates does not show updates correctly
From version 2.14.0 the `display-parent-updates` does not show newer parent versions anymore.

Given I have a project which has a newer parent version in our maven repository.
When I run `mvn org.codehaus.mojo:versions-maven-plugin:2.14.0:display-parent-updates`
Then I expect to see `The parent project has a newer version`

But instead I see `The parent project is the latest version`

Running with a plugin version of 2.13.0 it gives the correct result.

ajarmoniuk:
- Version range should not be specified in the retrieval at all; removed.
- If provided, the actual version range restriction should be used, and not the result of the toString() method
- Also, allowSnapshots was ignored for the mojo; corrected

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
@@ -53,6 +53,7 @@ import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.artifact.resolver.ArtifactResolutionException;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
+import org.apache.maven.artifact.versioning.Restriction;
 import org.apache.maven.artifact.versioning.VersionRange;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Dependency;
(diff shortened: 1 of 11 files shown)


ajarmoniuk:
Also corrected the way how range was interpreted if also a concrete version was passed inside the VersionRange object.

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -184,7 +185,12 @@ public ArtifactVersions lookupArtifactVersions(
                                     mavenSession.getRepositorySession(),
                                     new VersionRangeRequest(
                                             toArtifact(artifact)
-                                                    .setVersion(versionRange != null ? versionRange.toString() : "(,)"),
+                                                    .setVersion(ofNullable(versionRange)
+                                                            .map(VersionRange::getRestrictions)
```
ajarmoniuk, line 189: Here we're using the restriction present in the Version range object **if it's present** instead of the concrete version.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayParentUpdatesMojo.java
```diff
@@ -71,26 +69,17 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
                         .withArtifactId(getProject().getParent().getArtifactId())
-                        .withVersion(version)
+                        .withVersion(currentVersion)
                         .withType("pom")
                         .build());
 
         ArtifactVersion artifactVersion;
         try {
-            artifactVersion = findLatestVersion(artifact, versionRange, null, false);
+            artifactVersion = findLatestVersion(artifact, null, allowSnapshots, false);
```
ajarmoniuk, line 82: We're never actually interested in range-restricted versions; removing.

PR #898, opened 2023-01-03
Use ComparableVersion from Maven and remove the duplicate

ajarmoniuk:
Removing the duplicated ComparableVersion, which is also present in Maven. 

Because it's removed, I had to rework BoundArtifactVersion a bit. Now it's not "hacking" ComparableVersion anymore, ie, it's no longer using its private implementation to compute compareTo(). Instead, it's construction "infinity-like" version by using Integer.MAX_INT and uses this to do the comparison on the standard DefaultArtifactVersion.

BoundArtifactVersion is also officially, contractually immutable, and because of that, its `parseVersion` method throws an exception and is deprecated.

Also, because the ComparableVersion is gone, the cache is also gone, so I've implemented our own, but went a bit further: we're caching DefaultArtifactVersion instead, since it's also (almost) immutable, and we're not using it's only mutating method `parseVersion()` anywhere.

On this condition, it's safe to use the cache.


ajarmoniuk:

versions-common/src/main/java/org/codehaus/mojo/versions/ordering/BoundArtifactVersion.java
```diff
@@ -1,44 +1,106 @@
...
-    private final BoundComparableVersion comparator;
+    private final ArtifactVersion comparable;
 
     /**
-     * Constructs the instance
+     * Constructs the instance given the version in a text format.
+     * @param artifactVersion version in a text format
+     * @param segment most major segment that can change, i.e. <em>not</em> held in place
+     */
+    public BoundArtifactVersion(String artifactVersion, Segment segment) {
```
ajarmoniuk, line 42: So this is the key method, I guess... 

Hope this is clear, looks a bit like first year's comp-sci exercise, but we're simply parsing the version and splitting it to tokens upon:
- a change from numbers to letters or vice-versa
- an - character
- a dot

The new constructed version will have Inetger.MAX_VALUE on all places less than the fixed segment. Also, all non-strandard tokens will be replaced by zeros (like in the old impl).

ajarmoniuk:

Issue #74, opened 2016-05-09
Support for reporting new extenions versions


obfischer:
The Maven Versions Plugin already supports the reporting of new plugin versions (`versions:plugin-updates-report`) and new versions of dependencies (`versions:dependency-updates-report`). 

If not already supported, please enhence the Maven Versions Plugin also to report new versions of Maven extensions.

obfischer:
I would like to provide a patch for this feature request.

famod:
@obfischer have you ever gotten round to implementing this?

delanym:
@famod doesn't look like it https://github.com/obfischer/fork-of-versions-maven-plugin

50 open pull requests :(

ajarmoniuk:
Okay

ajarmoniuk:
Will do this in stages. First - `display-extension-updates`.

theit:
@ajarmoniuk Are there any news on this issue? I've seen your commits, and they look very promising (at least to me) :-)

ajarmoniuk:
No news, the PR is waiting to be merged.

theit:
Is there help needed with something?

ajarmoniuk:
I don't think so. From what I can gather, Sławek (the main maintainer) is currently quite busy with Maven 4-related issues.

theit:
Ok, thanks. Have you seen my [comment](https://github.com/ajarmoniuk/versions-maven-plugin/commit/1e4bd55238e686ee6757de39ffd2ba21c1febe23#r109271112) on your latest commit?

ajarmoniuk:
Can't see any comment. Let me check...

PR #908, opened 2023-01-24
Issue #74: Add display-extension-updates

ajarmoniuk:
First stage of extensions support.

*Not* supporting core extensions. Only build extensions are supported.


versions-common/src/main/java/org/codehaus/mojo/versions/utils/ExtensionBuilder.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
slawekjaranowski, line 2: It is not ASF project

slawekjaranowski (commented):
Extension can also by defined in `.mvn/extensions.xml`
https://maven.apache.org/configure.html#mvn-extensions-xml-file

ajarmoniuk:
@slawekjaranowski added support for core extensions declared in `.mvn/extensions`

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayExtensionUpdatesMojo.java
```diff
@@ -0,0 +1,298 @@
...
+import org.codehaus.mojo.versions.utils.SegmentUtils;
+import org.codehaus.plexus.util.StringUtils;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+
+import static java.util.Optional.of;
+import static org.codehaus.mojo.versions.api.Segment.MAJOR;
+
+/**
+ * Displays all build extensions that have newer versions available.
+ * <b>Does not check the core extensions.</b>
```
slawekjaranowski, line 61: Is it true?

slawekjaranowski (changes requested):

versions-model-core-extensions/src/main/mdo/core-extensions.mdo
```diff
@@ -0,0 +1,122 @@
...
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+
+-->
+
+<model xmlns="http://codehaus-plexus.github.io/MODELLO/1.4.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://codehaus-plexus.github.io/MODELLO/1.4.0 http://codehaus-plexus.github.io/modello/xsd/modello-1.4.0.xsd"
```
slawekjaranowski, line 24: There is 2.0.0 version of modelo xsd
https://codehaus-plexus.github.io/modello/modello.html

versions-maven-plugin/src/it/it-display-extension-updates-004/child/.mvn/extensions.xml
```diff
@@ -0,0 +1,8 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
```
slawekjaranowski, line 2: Core extensions can be defined only in project root - in child module it will be skipped as I know.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayExtensionUpdatesMojo.java
```diff
@@ -0,0 +1,293 @@
...
+import static java.util.Optional.of;
+import static org.codehaus.mojo.versions.api.Segment.MAJOR;
+
+/**
+ * Displays all build and core extensions that have newer versions available.
+ *
+ * @author Andrzej Jarmoniuk
+ * @since 2.15.0
+ */
+@Mojo(name = "display-extension-updates", threadSafe = true)
```
slawekjaranowski, line 61: Maybe we should use `aggregator` for such mojo
core extensions can be defined only in project root.

slawekjaranowski:
@ajarmoniuk please rebase this PR ... now I see in diff changes from master ...

ajarmoniuk:
> @ajarmoniuk please rebase this PR ... now I see in diff changes from master ...

Rebased.

versions-maven-plugin/src/it/it-display-extension-updates-004/child/.mvn/extensions.xml
slawekjaranowski: I don't see a check for `dummy-impl` so why this file exists?

slawekjaranowski (changes requested):
Ok, I see 
you create new module `versions-model-core-extensions` with contain a copy file core-extensions.mdo from Maven core.

I think that better will be use `org.apache.maven.cli.internal.extension.model.io.xpp3.CoreExtensionsXpp3Reader` even it is internal

ajarmoniuk:
> Ok, I see you create new module `versions-model-core-extensions` with contain a copy file core-extensions.mdo from Maven core.
> 
> I think that better will be use `org.apache.maven.cli.internal.extension.model.io.xpp3.CoreExtensionsXpp3Reader` even it is internal

Are you sure?

I asked you before when you requested me to change that:

>     Ok, so I have a question about this. Maybe I could appeal to the Maven project to make this public, or at least export the mdo?

>     I could also include the mdo file in the versions plugin. But what about its license then? > It's the ASF -- should we keep it as ASF then?

>     As an alternative, I could create a parser from scratch, be it manually, or using a bespoke XSD file and generate a parser from it.

> So, what do you think should be the course of action here?

So, what should it be?

ajarmoniuk:
> I don't see a check for dummy-impl so why this file exists?

to test if the .mvn/extensions is *not* processed in the child.

ajarmoniuk:
@slawekjaranowski and if you need time to decide, feel free to remove this from the upcoming release. 

It's not so easy. 

Maven Embedder 3.2.5 does not have Core Extensions parser. Only 3.3.9 and later has it. So we cannot use the one from Maven (so, it can't be _provided_).

slawekjaranowski:
> @slawekjaranowski and if you need time to decide, feel free to remove this from the upcoming release.
> 
> It's not so easy.
> 
> Maven Embedder 3.2.5 does not have the Core Extensions parser. Only 3.3.9 and later has it. So we cannot use the one from the same version as Maven.
> 
> https://maven.apache.org/ref/3.2.5/maven-embedder/xref/index.html

Ok, I must review of it more carefully ... so I will do it again in next week

Maybe some one else want to look also - @slachiewicz ?

ajarmoniuk:
@slawekjaranowski @slachiewicz 

It's not the only case like this in this plugin. Please have a look at e.g. parsing raw pom. Currently it's an ad-hoc parser created from scratch in the plugin, but it actually depends on Maven and the underlying document, which is **owned** by the Maven project. So it's a very similar case.

ajarmoniuk:
@slawekjaranowski I propose that we keep a version of this PR with the file from the Apache Maven project with its own ASF license and its own Modello version. Until there comes an API from Maven, which we can request or implement.

versions-model-core-extensions/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+            <goals>
+              <goal>xpp3-reader</goal>
+              <goal>java</goal>
+            </goals>
+            <phase>generate-sources</phase>
+          </execution>
+          <execution>
+            <id>generate-site-doc</id>
+            <goals>
+              <goal>xdoc</goal>
```
slawekjaranowski, line 44: We should not generate xsd for this model - it is internally used for reading extensions.xml

slawekjaranowski (commented):
I think that we not need new module only for a few classes, we not need xsd for it, it is only for internal purpose.

I would like to put core-extensions.mdo into versions-common where is only used.

ajarmoniuk:
> I think that we not need new module only for a few classes, we not need xsd for it, it is only for internal purpose.
> 
> I would like to put core-extensions.mdo into versions-common where is only used.

Applied.

ajarmoniuk:
> I think that we not need new module only for a few classes, we not need xsd for it, it is only for internal purpose.
> 
> I would like to put core-extensions.mdo into versions-common where is only used.

Oh yes, I remember it now. The reason why I created that module was that I couldn't go around javadoc generation errors -- classes were generated.

ajarmoniuk:
@slawekjaranowski is there anything else to be done about this PR?

slawekjaranowski:
Hi,
I was removed `maven-embedder` from dependency - we have a copy
I restored license header in files with longer history - we don't know why such license was used ... 

Wait for build and we can go.

PR #924, opened 2023-02-25
Resolves #916: Partially reverted #799.

### LINKED ISSUE #916: versions:set doesn't set version if non-direct properties are used in version element
Maven supports 5 property classes in pom.xml files:
1. `env.X` - environment variables
2. `project.x` - any project properties from the pom.xml
3. `settings.x` - properties from settings.xml
4. Java Systems Properties, e.g. `java.home`
5. `x` - properties from `<properties>` section of the pom.xml

If any of the 1-4 property classes is used in `version` element (even transitively) the resolution of the current version will fail, and no new version will be set.

E.g. parent pom.xml
```
<project>

    <groupId>com.example.app</groupId>
    <artifactId>my-parent</artifactId>
    <version>1.0.0</version>
    <packaging>pom</packaging>

</project>
```
child pom.xml
```
<project>

    <parent>
        <groupId>com.example.app</groupId>
        <artifactId>my-parent</artifactId>
        <version>1.0.0</version>
    </parent>

    <groupId>com.example.app</groupId>
    <artifactId>my-app</artifactId>
    <version>${project.parent.version}</version>
</project>

```
Attempt to set version:
```
$ mvn versions:set -DnewVersion=1.1.1
(...)
[INFO] Building my-app 1.0.0
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- versions-maven-plugin:2.14.2:set (default-cli) @ my-app ---
[INFO] Searching for local aggregator root...
[INFO] Local aggregation root: C:\Work\projects\3rd_party\tst
[INFO] Processing change of com.example.app:my-app:1.0.0 -> 1.1.1
expression: project.parent.version no value
```

This caught us off guard - we use versions:set to override the version only for our 'dev' builds which are rather seldom run. 

Workaround
```
$ mvn versions:set -DnewVersion=1.1.1 -DoldVersion='*'
(...)
[INFO] Building my-app 1.0.0
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- versions-maven-plugin:2.14.2:set (default-cli) @ my-app ---
[INFO] Searching for local aggregator root...
[INFO] Local aggregation root: C:\Work\projects\3rd_party\tst
[INFO] Processing change of com.example.app:my-app:* -> 1.1.1
expression: project.parent.version no value
[INFO] Processing com.example.app:my-app
[INFO]     Updating project com.example.app:my-app
[INFO]         from version  to 1.1.1
```

I know this might be a corner case, and not really an issue, since it is encouraged that the version element is constant and not an expression. Furthermore I couldn't come up with other probable use cases.

Having said that, I think that because some property interpolation is already handled by the plugin, one might expect that all property classes may be used, and be surprised by this behavior.

Please consider whether it's worth handling.

ajarmoniuk:
- **Improved documentation**
- Reviewed the code and the semantics of processAllModules, groupId, artifactId, oldVersion params
- Dropped support for partially interpolating the properties to be matched with groupId:artifactId:oldVersion
- Matching occurs always unless *processAllModules* is set
- If *processAllModules* is not set, *oldVersion* will be respected

### Conclusion

It basically turns out that in order to fully support properties, we would have to build the project for every module and construct all property values, which is overkill for the purpose of selecting the modules to process. So I decided to drop support for interpolating properties, just like it was before 2.14.0.

Instead, we're matching against uniterpolated values, which again is optional. Old behaviour is retained thanks to applying the uninterpolated values from project by default.


ajarmoniuk:
@slachiewicz please review

versions-maven-plugin/pom.xml
```diff
@@ -289,6 +289,9 @@
                 <configuration>
                   <streamLogs>false</streamLogs>
                   <debug>true</debug>
+                  <environmentVariables>
+                    <!--                    <MAVEN_DEBUG_OPTS>-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=localhost:8000</MAVEN_DEBUG_OPTS>-->
+                  </environmentVariables>
```
slawekjaranowski, line 294: ?

Issue #299, opened 2018-08-20
allow* parameters not working on display-dependency-updates


meeque:
I'm trying to use the `versions:display-dependency-updates` goal from the CLI, but forbid _major_ updates. I tried both of the following:
```
mvn versions:display-dependency-updates -DallowAnyUpdates=false -DallowMajorUpdates=false
mvn versions:display-dependency-updates -DallowAnyUpdates=false -DallowMajorUpdates=false -DallowMinorUpdates=true -DallowIncrementalUpdates=true -DallowSnapshots=false
```

In any case, I'm still seeing outputs like the following:
```
[INFO]   org.springframework:spring-core .......... 4.3.16.RELEASE -> 5.0.0.RC4
```
As you can see, the major version changed from 4 to 5, which I would not expect. Instead, was hoping for the following (based on today's versions on Maven Central):
```
[INFO]   org.springframework:spring-core ..... 4.3.16.RELEASE -> 4.3.18.RELEASE
```

Sorry, if I'm missing something fundamental, or have a typo in the above. I tried hard to get it right, and rule everything out. 

Btw, the above behavior is not limited to that specific dependency. Other dependencies are affected, too. Could this be caused by the weird version suffixes that projects such as Spring and Netty are using?

Fyi, some version information:
```
[INFO] --- versions-maven-plugin:2.5:display-dependency-updates (default-cli) @ Xyz ---
```
```
Apache Maven 3.5.4 (1edded0938998edf8bf061f1ceb3cfdeccf443fe; 2018-06-17T20:33:14+02:00)
Maven home: /usr/local/Cellar/maven/3.5.4/libexec
Java version: 1.8.0_181, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.6", arch: "x86_64", family: "mac"
```

dougbreaux:
2.5 years on, relatively new to Maven and really wanting & expecting this to work. And the documentation certainly says that it should.

meeque:
Thanks for resurfacing this one, @dougbreaux. Frankly, I have accepted that this just does not work as expected a long time ago. In fact, I had forgotten about this completely.

It would still be great to hear feedback from a contributor. Is this considered a bug? Or is it expected behavior? How do we need to adjust our expectations?

[...]

PR #941, opened 2023-04-09
#299: allowAnyUpdates should be ignored

ajarmoniuk:
This might be seen as a breaking change, but I think it is in line with what most users expect to happen anyway.

So, this change sanctions these expectations -- allowAnyUpdates shall effectively assume the value of `false` if any of the other modifies are set to `false`.

Actually, this means that it's redundant, but the fact is that it's getting removed in version 3.0 anyway.

@slawekjaranowski what is your opinion?

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
@@ -21,13 +21,7 @@ package org.codehaus.mojo.versions;
 
 import javax.inject.Inject;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.*;
 
 import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
@@ -54,9 +48,7 @@ import static java.util.Optional.of;
 import static org.apache.commons.lang3.StringUtils.countMatches;
 import static org.codehaus.mojo.versions.api.Segment.MAJOR;
 import static org.codehaus.mojo.versions.filtering.DependencyFilter.filterDependencies;
-import static org.codehaus.mojo.versions.utils.MavenProjectUtils.extractDependenciesFromDependencyManagement;
-import static org.codehaus.mojo.versions.utils.MavenProjectUtils.extractDependenciesFromPlugins;
-import static org.codehaus.mojo.versions.utils.MavenProjectUtils.extractPluginDependenciesFromPluginsInPluginManagement;
+import static org.codehaus.mojo.versions.utils.MavenProjectUtils.*;
 
 /**
  * Displays all dependencies that have newer versions available.
@@ -246,11 +238,14 @@ public class DisplayDependencyUpdatesMojo extends AbstractVersionsDisplayMojo {
     private boolean allowIncrementalUpdates = true;
 
     /**
-     * Whether to allow any version change to be allowed. This keeps
-     * compatibility with previous versions of the plugin.
-     * If you set this to false you can control changes in version
-     * number by {@link #allowMajorUpdates}, {@link #allowMinorUpdates} or
-     * {@link #allowIncrementalUpdates}.
+     * <p>Ignored -- largely replaced by {@linkplain #allowMajorUpdates},
+     * {@linkplain #allowMinorUpdates} and {@linkplain #allowIncrementalUpdates},
+     * which are equal to {@code true} by default.</p>
+     *
+     * <p><b>Please note: Prior to version 2.16.0, leaving this parameter at its default
+     * value ({@code true}) would mean that the plugin would <u>ignore</u>
+     * {@linkplain #allowMajorUpdates}, {@linkplain #allowMinorUpdates}, and {@linkplain #allowIncrementalUpdates},
+     * which confused many users.</b></p>
      *
      * @since 2.5
      * @deprecated This will be removed with version 3.0.0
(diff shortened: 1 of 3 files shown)


ajarmoniuk:
Looks like this is also in line with what users expect – see the latest comments in #299.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -489,6 +484,11 @@ protected void validateInput() throws MojoExecutionException {
         validateGAVList(pluginDependencyExcludes, 3, "pluginDependencyExcludes");
         validateGAVList(pluginManagementDependencyIncludes, 3, "pluginManagementDependencyIncludes");
         validateGAVList(pluginManagementDependencyExcludes, 3, "pluginManagementDependencyExcludes");
+        if (getLog() != null
+                && allowAnyUpdates
+                && !(allowMajorUpdates && allowMinorUpdates && allowIncrementalUpdates)) {
+            getLog().warn("Assuming allowAnyUpdates false because one or more other \"allow\" switches is false.");
```
slawekjaranowski, line 490: `allowAnyUpdates` has default value as `true` so when user provide only one `allow*` switch this warning will be always displayed.
As now we ignore at all `allowAnyUpdates` maybe it is not needed.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -246,11 +238,14 @@ public class DisplayDependencyUpdatesMojo extends AbstractVersionsDisplayMojo {
...
-     * number by {@link #allowMajorUpdates}, {@link #allowMinorUpdates} or
-     * {@link #allowIncrementalUpdates}.
+     * <p>Ignored -- largely replaced by {@linkplain #allowMajorUpdates},
+     * {@linkplain #allowMinorUpdates} and {@linkplain #allowIncrementalUpdates},
+     * which are equal to {@code true} by default.</p>
+     *
+     * <p><b>Please note: Prior to version 2.16.0, leaving this parameter at its default
+     * value ({@code true}) would mean that the plugin would <u>ignore</u>
+     * {@linkplain #allowMajorUpdates}, {@linkplain #allowMinorUpdates}, and {@linkplain #allowIncrementalUpdates},
+     * which confused many users.</b></p>
```
slawekjaranowski, line 248: I prefer a documentation for current version - and current behavior.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
ajarmoniuk: > As now we ignore at all allowAnyUpdates maybe it is not needed.

The problem is that we do not ignore `allowAnyUpdates` -- we ignore all the other `allow` options if `allowAnyUpdates` is `true`.

See: https://github.com/mojohaus/versions/blob/0fe821e400fe917b5fc342fe5712071dd0490473/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java#L514

That is basically the reason why I created this PR.

ajarmoniuk: In other words, if the user only sets `allowMajorUpdates=false`, that option will be ignored because the user also must use `allowAnyUpdates=false` together with that option for it to work... Which is flawed in my opinion.

slawekjaranowski: I see that after this change `allowAnyUpdates` will be ignored - so it will be not important what user set - right?

ajarmoniuk: Yes, with a warning message.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -506,7 +506,7 @@ static void validateGAVList(List<String> gavList, int numSections, String argume
     }
 
     private Optional<Segment> calculateUpdateScope() {
-        return allowAnyUpdates
+        return allowMajorUpdates && allowMinorUpdates && allowIncrementalUpdates
```
slawekjaranowski, line 509: The similar logic is in method  `SegmentUtils#determineUnchangedSegment` in this place looks  like complicated .... but should be refactored in separated PR - if possible.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
slawekjaranowski: ok, but settings or not `allowAnyUpdates` will mo have any  effect only warning message will be or not displayed
So I think that warning is such case is redundant
I don't see any more usage of `allowAnyUpdates` ...

ajarmoniuk: I don't see it either.

Ideally, I would remove it.

slawekjaranowski: Do it. `allowAnyUpdates` is only used in this goal. Will be more clear.

slawekjaranowski: agree - we can wait a few days.
I would to release next version at next week - about 25 - 26 of May

ajarmoniuk:
While working on the refactoring job, I've discovered a bug around `allowAnyUpdates`: if `allowAnyUpdates` is `false` and all other allows are `true`, the result differs from when `allowAnyUpdates` is `true`.

In the first case, the `updateScope` is `Optional.of(MAJOR)`, in the second case (with allowAnyUpdates), it's `Optional.empty()`. 

Strange that we haven't found it with our tests. Well, I'm fixing this now (we'll get rid of `allowAnyUpdates`), but this will take some time.

slawekjaranowski:
Looks more complicated ... so I have proposition to postpone this and related issues to next release.

ajarmoniuk:
Actually, I'd close this PR and I would postpone #961.

On the other hand, #960 is important and almost ready.

ajarmoniuk:
> On the other hand, https://github.com/mojohaus/versions/issues/960 is important and almost ready.

Not ready, I see that integration tests are failing, so there's more work needed. Also, I need to see if this is restricted to display-dependency-updates. So, let's postponse that one too.

PR #948, opened 2023-04-29
Require Maven 3.5.4+

slachiewicz:


slawekjaranowski:
README to fix

.github/workflows/maven.yml
```diff
@@ -26,6 +26,7 @@ jobs:
     with:
       ff-goal: '-P run-its install'         # site use project version for reporting
       verify-goal: '-P run-its install'     # should be the same as for first build
+      maven-matrix: '[ "3.5.4", "3.9.1" ]'  # Maven versions matrix for verify builds
```
slawekjaranowski, line 29: build will take about 1 hour 😄

slachiewicz: yes, but at last I would like to test somehow once

ajarmoniuk:
Currently, version also needs to be updated here: https://github.com/mojohaus/versions/blob/6cd759ce2b1e884ebbbf3dece70b156d1b102753/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayPluginUpdatesMojo.java#L128

However, I think that this assumption is bogus: minimum maven version of the POM doesn't have to be the same as the minimum version required to run versions-maven-plugin.

PR #965, opened 2023-05-26
Alter session for version discovery

cstamas:
Alter session used for discovery, pinpoint ONLY to processed artifact to update ALWAYS. This would fix all the problems reported here https://github.com/mojohaus/versions/issues/959 and here https://issues.apache.org/jira/browse/MRESOLVER-363 just like `-U` does (and people do it by reflex). But this is more correct way, as `-U` is global and refreshes too much.

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
@@ -274,6 +274,7 @@ public class DefaultVersionsHelper implements VersionsHelper {
                 repositories = emptyList();
             }
 
+            // in future sessions may become AutoCloseable, so all this would go into try-with-resource block
             DefaultRepositorySystemSession versionDiscoverSession =
                     new DefaultRepositorySystemSession(mavenSession.getRepositorySession());
             versionDiscoverSession.setUpdatePolicy(RepositoryPolicy.UPDATE_POLICY_ALWAYS);
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
@@ -83,6 +83,7 @@ import org.codehaus.plexus.component.configurator.expression.ExpressionEvaluatio
 import org.codehaus.plexus.component.configurator.expression.ExpressionEvaluator;
 import org.codehaus.plexus.util.StringUtils;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+import org.eclipse.aether.DefaultRepositorySystemSession;
 import org.eclipse.aether.repository.AuthenticationContext;
 import org.eclipse.aether.repository.RemoteRepository;
 import org.eclipse.aether.repository.RepositoryPolicy;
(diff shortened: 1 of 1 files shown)


cstamas:
@slachiewicz @ajarmoniuk ping

ajarmoniuk:
I wish I were a maintainer, but for the time being I can't approve the workflow 😉

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
```diff
@@ -273,11 +274,15 @@ public ArtifactVersions lookupArtifactVersions(
                 repositories = emptyList();
             }
 
+            // in future sessions may become AutoCloseable, so all this would go into try-with-resource block
```
michael-o, line 277: This should be marked with TODO then

michael-o (changes requested):

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
cstamas: unsure, i just left it here. IMHO is not TODO as session is NOT YET AutoCloseable, may be 1.11.x or maybe 2.0 but nobody can tell WHEN it will be. OTOH, "TODO" would mean "do it if you can", but it cannot be done today.

cstamas:
Am not a maintainer either here, so I will just leave this here for @slawekjaranowski or anyone else to handle. Comment may be removed, I just added it like "head up", but it may happen in near or far future (I cannot tell).

versions-common/src/main/java/org/codehaus/mojo/versions/api/DefaultVersionsHelper.java
michael-o: NOTE?

slawekjaranowski:
I'm not sure if it is good way, after it every execute plugin will cause download every metadata again.

For projects with many dependencies  it can be an issue.

cstamas:
@slawekjaranowski every metadata? only targeted, only `artifact` that is in call. Maybe vrsions plugin could control, and have "own" `-U` instead? 

Point with `mvn -U` is that it is global, so will download "every metadata maven touches" (so literally EVERY in build), while this change downloads ONLY the metadata for `artifact`

cstamas:
@slawekjaranowski or if I misunderstood you, and you mean "will refresh metadata for maven-clean-plugin in EVERY module of the possibly 100 module build? so perform 100 refreshes?" That is not the case, as resolver refreshes once per session, so yes, it will refresh once in first module clean plugin is found, and from then on (in remaining 99 modules) cached results are used.

See https://github.com/apache/maven-resolver/blob/maven-resolver-1.9.10/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/DefaultUpdateCheckManager.java#L267

slawekjaranowski:
As I see issue from #959 is that we don't have or is wrong property for last modified in properties file.
According to https://github.com/apache/maven-resolver/pull/292/files#diff-51ca9aed78d3713d9818b753db3b9e9629743a7e773292e686f93950fca4c2b0R103 in such case default policy for repository is used.

Because default central plugin repository has policy never - https://github.com/apache/maven/blob/6f136ef4d2b4938e939189e9d2e3e0faa11a20e9/maven-model-builder/src/main/resources/org/apache/maven/model/pom-4.0.0.xml#L53
we try to override policy for whole session for all repositories.

Last I did improvement for repository with never policy to use daily #957 - maybe it will be good enough is such case.

cstamas:
Cool then, was unaware that this plugin overrides _configured_ policy.
I am still convinced, that is PR (conditionally, like some `-Dforce` flag) would be useful for plugin (as aside of daily, user has no other means to say "update  now!" than `mvn -U` which is too much).

But am fine, so if it works as you tested, am fine with it.

PR #966, opened 2023-05-27
Resolves: display-dependency-managment only shows updates from the most major allowed segment

ajarmoniuk:
Changed the way display-dependency-updates checks for updates. It turns out that other plugins did not use the same method, so they are safe. plugin- and parent-updates don't even offer these "allow-" parameters.

DisplayDependencyUpdates used the same routine that dependency reports did, but that routine only selected updates within a given segment -- that's why I've split the functions in the API between `restrictionForUnchangedSegment`, `restrictionForSelectedSegment`, and `restrictionForIgnoreScope`. Perhaps there's a way to unite those, but that's for a future release.

Also, I've discovered another bug, but with a less significance because it only affects version ranges.

If an artifact is defined using two or more ranges, and the selected version falls in the first range, the versions in between the ranges will never get selected.

```
[, 1.0.0]   v1=1.1.0   v2=1.1.1   v3=1.9.0  (2.0.0, 10.0.0)
```

from the above, v1, v2, v3 will never get shown as possible updates.

That is fixed for display-dependency-updates. Not sure about the other mojo's as they use different ways of retrieving updates.

Overall it's a big mess and badly needs rearchitecting -- but that's not something for this while.


ajarmoniuk:
also updated the copyright in modified files

slawekjaranowski:
> also updated the copyright in modified files

We should not change license header in existing files ... theoretically we should dig in history and confirm with authors of each file

ajarmoniuk:
> > also updated the copyright in modified files
> 
> We should not change license header in existing files ... theoretically we should dig in history and confirm with authors of each file

Reverted those

PR #974, opened 2023-06-09
Resolves #973: NPE if actual version is null for a dependency

### LINKED ISSUE #973: org.codehaus.mojo:versions-maven-plugin:2.16.0:display-dependency-updates failed: Cannot invoke "Object.toString()" because "otherVersion" is null
The project for which the plugin was called: https://github.com/publicLibs/RollingMavenHeader/commit/0a93dbce4c6657f816362971654d862d9798d510

command
```sh
mvn versions:display-dependency-updates -X > FULL_ERROR_X.txt
```
[FULL_ERROR_X.txt](https://github.com/mojohaus/versions/files/11699378/FULL_ERROR_X.txt)

ajarmoniuk:
In case a dependency version is specified in dependencyManagement, a dependency can be versionless.

Dependency updates goals would then attempt to find an updated version to a versionless dependency, which would fail with an NPE or, if that is prevented, an attempt would have been made to find an update to version specified as [,0], which would be any version.

Preventing both issues.

### DIFF
--- a/versions-maven-plugin/src/it/it-display-dependency-updates-issue-973-versionless/pom.xml
+++ b/versions-maven-plugin/src/it/it-display-dependency-updates-issue-973-versionless/pom.xml
@@ -7,6 +7,12 @@
   <packaging>pom</packaging>
   <name>display-dependency-updates-from-plugins</name>
 
+  <description>
+    Edge case: the pom.xml will fail when we try executing the actual plugin
+    as it is invalid -- lacks version, which caused an NPE (Issue #973)
+    when executing display-dependency-updates.
+  </description>
+
   <dependencyManagement>
     <dependencies>
       <dependency>
--- a/versions-maven-plugin/src/it/it-display-dependency-updates-issue-973-versionless/pom.xml
+++ b/versions-maven-plugin/src/it/it-display-dependency-updates-issue-973-versionless/pom.xml
@@ -27,8 +27,8 @@
     <pluginManagement>
       <plugins>
         <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-clean-plugin</artifactId>
+          <groupId>localhost</groupId>
+          <artifactId>dummy-maven-plugin</artifactId>
           <dependencies>
             <dependency>
               <groupId>localhost</groupId>
--- a/versions-maven-plugin/src/it/it-display-dependency-updates-issue-973-versionless/pom.xml
+++ b/versions-maven-plugin/src/it/it-display-dependency-updates-issue-973-versionless/pom.xml
@@ -1,0 +1,37 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>localhost</groupId>
+  <artifactId>it-display-dependency-updates-002</artifactId>
+  <version>1.0</version>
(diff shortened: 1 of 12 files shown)


versions-maven-plugin/src/it/it-display-dependency-updates-issue-973-versionless/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+      <plugins>
+        <plugin>
+          <groupId>localhost</groupId>
+          <artifactId>dummy-maven-plugin</artifactId>
+          <dependencies>
+            <dependency>
+              <groupId>localhost</groupId>
+              <artifactId>dummy-api</artifactId>
+            </dependency>
+          </dependencies>
```
slawekjaranowski, line 31: I think that such example is wrong ... it is not possible to execute such project
version in plugin/dependencies must be set and it is not resolved from project dependencies/depndcyManagemnt

ajarmoniuk: This example is the only case I've found that reproduces the issue. Yes, it does work with Maven.

slawekjaranowski: please look at my commit and build status

slawekjaranowski:
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'build.plugins.plugin[org.apache.maven.plugins:maven-clean-plugin].dependencies.dependency.version' for localhost:dummy-api:jar is missing. @ line 27, column 25
```

ajarmoniuk: Strange. I don't understand why it worked with my commit then. With my pom.xml, the issue was being hit and the fix did work. 

So now it remains to be found how we could possibly reproduce the issue.

slawekjaranowski: In your commit plugin was not executed - only versions-plugin try to analyze project

ajarmoniuk:
The plugin itself must define a dependency version in its dependencies or dependency management. 

This does trigger the error and will be helped with the patch. 

For this to work though as an it, the plugin needs to be executed.

```xml
    <pluginManagement>
      <plugins>
        <plugin>
          <groupId>com.github.spotbugs</groupId>
          <artifactId>spotbugs-maven-plugin</artifactId>
          <version>4.7.3.4</version>
          <dependencies>
            <!-- overwrite dependency on spotbugs if you want to
            specify the version of spotbugs -->
            <dependency>
              <groupId>com.github.spotbugs</groupId>
              <artifactId>spotbugs</artifactId>
            </dependency>
          </dependencies>
        </plugin>
      </plugins>
    </pluginManagement>
```

ajarmoniuk:
@slawekjaranowski Ok, so I guess this is another edge case: the pom.xml will fail when we try executing the actual plugin as it is invalid -- lacks version. However, this **faulty** pom.xml will cause an NPE **on the versions plugin** if we try bumping the versions.

slawekjaranowski:
> @slawekjaranowski Ok, so I guess this is another edge case: the pom.xml will fail when we try executing the actual plugin as it is invalid -- lacks version. However, this **faulty** pom.xml will cause an NPE **on the versions plugin** if we try bumping the versions. So, maybe let's allow this faulty pom.xml as the goal here is not to execute the plugin, but to make the versions plugin not fail while processing the pom.

Sounds reasonable - we only need comments on IT tah such configurations is not working

slawekjaranowski:
Dropped my test commit 😄

ajarmoniuk:
Added a description. Did not squash :)

PR #994, opened 2023-07-29
Resolves #990: Check prerequisites if required enforcer Maven version is empty

### LINKED ISSUE #990: Bug: prerequisites are not detected
When doing `versions:display-plugin-updates` on this module https://github.com/maveniverse/mima/tree/75a06a1132103506614b33a2cba2229923e86260/demo/library-maven-plugin

I get this message:
```
[INFO] --------< eu.maveniverse.maven.mima.demo:library-maven-plugin >---------
[INFO] Building eu.maveniverse.maven.mima.demo:library-maven-plugin 1.0.0-SNAPSHOT [5/5]
[INFO]   from library-maven-plugin/pom.xml
[INFO] ----------------------------[ maven-plugin ]----------------------------
[INFO] 
[INFO] --- versions:2.16.0:display-plugin-updates (default-cli) @ library-maven-plugin ---
[INFO] 
[INFO] 
[INFO] 
[WARNING] Project does not define minimum Maven version required for build, default is: 3.2.5
[INFO] Plugins require minimum Maven version of: 3.2.5
[INFO] 
[WARNING] Project (which is a Maven plugin) does not define required minimum version of Maven.
[WARNING] Update the pom.xml to contain
[WARNING]     <prerequisites>
[WARNING]       <maven><!-- minimum version of Maven that the plugin works with --></maven>
[WARNING]     </prerequisites>
[WARNING] To build this plugin you need at least Maven 3.2.5
[WARNING] A Maven Enforcer rule can be used to enforce this if you have not already set one up
[WARNING] See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
```

But https://github.com/maveniverse/mima/blob/75a06a1132103506614b33a2cba2229923e86260/demo/library-maven-plugin/pom.xml#L46-L48

While true, I have no enforcer to enforce build time Maven version.

Still, why does plugin claim I have no prerequisites, while I do?

ajarmoniuk:
Check prerequisites if enforcer check for Maven version is not present.

~~Prerequisites are checked after enforcer -- disputable, but at least this keeps the behavior backwards compatible.~~

The plugin will choose the version which is the highest between: default maven version, prerequisites, enforcer.

@slawekjaranowski please check

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/MinimalMavenBuildVersionFinder.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/MinimalMavenBuildVersionFinder.java
@@ -1,15 +1,21 @@
 package org.codehaus.mojo.versions;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.model.PluginExecution;
+import org.apache.maven.model.Prerequisites;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.project.MavenProject;
 import org.codehaus.mojo.versions.utils.DefaultArtifactVersionCache;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
+
+import static java.util.Optional.ofNullable;
 
 /**
  * Finds the minimal Maven version required to build a Maven project.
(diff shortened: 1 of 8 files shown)


ajarmoniuk:
Actually, I think it should take the greater of the two. I'll redo the PR.

Also, I don't really think that they're should be a default maven version. Currently it's 3.2.5, which is the minimum version required by the versions plugin, but the POM being analysed is executed without the versions plugin being involved and thus it's minimum version shouldn't be taken into account.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/MinimalMavenBuildVersionFinder.java
```diff
@@ -1,16 +1,18 @@
 package org.codehaus.mojo.versions;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
```
slawekjaranowski, line 3: 😄

ajarmoniuk: That' spotless apply :)

ajarmoniuk: Corrected.

PR #996, opened 2023-07-30
Resolves #995: Minimum version required for Versions Maven Plugin should is not the minimum version for the project

### LINKED ISSUE #995: DisplayPluginUpdates: default minimum maven version should not be assumed to be the same as the one for versions-maven-plugin
DisplayPluginUpdates will use the minimum version of Maven required to execute versions-maven-plugin to also be the minimum Maven version for the project that is being analysed. However, I see no reason why this should be the case. That would only be the case if the versions plugin is used in the POM, but obviously doesn't have to be the case.

In other words: we cannot make assumptions about the version required to execute the project based on the version that is currently being used to run versions-maven-plugin. Those can be different.

I think the assumption should be dropped.

ajarmoniuk:
We cannot make assumptions about the version required to execute the project based on the version that is currently in use to run versions-maven-plugin. Those can be different.

Because of that, I had to modify two integration tests.


slachiewicz (approved):

slawekjaranowski:
@ajarmoniuk please resolve conflict

slachiewicz (approved):

Issue #1002, opened 2023-09-08
Make processDependencyManagementTransitive as false by default


slawekjaranowski:
In project when we use import in dependency management, 
by default many updates are reported which is not possible to change in project.

I think that more usable is to know that new version of imported pom is available.

slawekjaranowski:
What do you think?

slawekjaranowski:
@slachiewicz @ajarmoniuk ?

ajarmoniuk:
+1

slachiewicz:
+1

slawekjaranowski:
Interesting exercise ... I raised an issues: #1004

Issue #1004, opened 2023-09-11
depManagement from parent should be used with processDependencyManagementTransitive=false


slawekjaranowski:
When we set `processDependencyManagementTransitive=false` dependencyManagement from parent should be still used.

eg. We have a IT - `it-dependency-updates-aggregate-report-issue-333` which will fail with `processDependencyManagementTransitive=false`

According to documentation:

```
Whether to process the dependencyManagement part transitive or not. 
In case of <type>pom</type>and <scope>import</scope> ...
```

I understood that only items `imported` by `pom` are skipped ... not everything from parent pom. 

First we need agree what this options means and next check if implementation meet requirements.

ajarmoniuk:
The documentation is wrong. 

If I didn't take the documentation of the parameter into account, it would look OK to me:

The dependencyManagement item missing from the aggregate report with `processDependencyManagementTransitive=false` is dummy-lib:2.0.0, which is a transitive dependencyManagement item from the parent of pom.xml -- see https://github.com/mojohaus/versions/blob/1d039561071cd3dffb7d9b98d689eee25cf6fd99/versions-maven-plugin/src/it/it-dependency-updates-aggregate-report-issue-333/pom.xml#L7

When debugging, I see that the functional difference is that with `processDependencyManagementTransitive=false`, we're taking dependency management items from the original model whereas with the parameter set to `true`, from the interpolated model. The latter will contain all transitive dependencies, not just with scope `pom` or `import`.

I don't see any reflection of the original documentation in the plugin code. I suggest that we change the docs to read that the option disables reporting on transitive dependency management, ie taken from parent or dependencies with scope `import`.

PR #1005, opened 2023-09-11
Make processDependencyManagementTransitive as false by default

slawekjaranowski:
Fi: #1002


slawekjaranowski:
First we need resolve - investigate: #1004

PR #1017, opened 2023-10-23
Align update-parent and display-parent-update

### LINKED ISSUE #1016: display-parent-updates should support allowMajorUpdates, allowMinorUpdates and related parameters
Since #966 (2.16.0) most `display-*-updates` and `update-*` goals consistently support a set of common parameters to narrow the scope of used dependencies (allowMajorUpdates, allowMinorUpdates, allowIncrementalUpdates, ...) However, **display-parent-updates** does not.

If you agree I'm happy to contribute a PR to make sure **display-parent-updates** accepts the same parameters as **update-parent**.

cachescrubber:
Resolves #1016


cachescrubber:
I added another Test, `DisplayParentUpdatesMojoOutputTest` which is based on `AbstractMojoTestCase` like all other Display*Tests. As soon as I add a parent to my test-cases pom, test execution will fail with

`java.lang.IllegalArgumentException: Invalid repository system session: LocalRepositoryManager is not set.`

ajarmoniuk:
Hi @cachescrubber, do you need any more assistance with the failing tests?

cachescrubber:
@ajarmoniuk  I gave on on the *Output test. Instead, I aligned `DisplayParentUpdatesMojoOutputTest` (new) with `UpdateParentMojoTest`. Could you please review my changes?

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayParentUpdatesMojo.java
```diff
@@ -46,6 +60,78 @@ public class DisplayParentUpdatesMojo extends AbstractVersionsDisplayMojo {
...
+    // ------------------------------ FIELDS ------------------------------
+
+    /**
+     * <p>If {@code skipResolution} is not set, specifies the <em>bottom</em> version considered
+     * for target version resolution. If it is a version range, the resolved version will be
+     * restricted by that range.</p>
+     *
+     * <p>If {@code skipResolution} is {@code true}, will specify the target version to which
+     * the parent artifact will be updated.</p>
+     * @since 2.16.1
```
ajarmoniuk, line 72: I guess that should be `2.17.0`. Ditto for the rest.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayParentUpdatesMojo.java
```diff
@@ -67,20 +153,13 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-        String currentVersion = getProject().getParent().getVersion();
-        Artifact artifact = getHelper()
-                .createDependencyArtifact(DependencyBuilder.newBuilder()
-                        .withGroupId(getProject().getParent().getGroupId())
-                        .withArtifactId(getProject().getParent().getArtifactId())
-                        .withVersion(currentVersion)
-                        .withType("pom")
-                        .build());
-
+        String currentVersion = parentVersion == null ? getProject().getParent().getVersion() : parentVersion;
```
ajarmoniuk, line 156: or you could use Optional.ofNullable(parentVersion).orElse(getProject().getParent().getVersion())

but that's not really important ;)

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayParentUpdatesMojo.java
```diff
@@ -122,6 +201,56 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         }
     }
 
+    protected ArtifactVersion resolveTargetVersion(String initialVersion)
```
ajarmoniuk, line 204: I'd move this method to `VersionHelper` (so we would also remove it from `UpdateParentMojo`).

ajarmoniuk (changes requested):

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayParentUpdatesMojo.java
cachescrubber: I would rather not do so, because `resolveTargetVersion` is expecting nullable String arguments anyway. But I renamed current to initialVersion, in order to be aligned with `UpdateParentMojo` .

ajarmoniuk: I meant, those two expressions are equivalent. Another idiom.

ajarmoniuk (approved):

ajarmoniuk:
@slawekjaranowski please approve the workflow and/or raise additional PR commments.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayParentUpdatesMojo.java
```diff
@@ -46,6 +60,78 @@ public class DisplayParentUpdatesMojo extends AbstractVersionsDisplayMojo {
...
+     * <p>If {@code skipResolution} is not set, specifies the <em>bottom</em> version considered
+     * for target version resolution. If it is a version range, the resolved version will be
+     * restricted by that range.</p>
+     *
+     * <p>If {@code skipResolution} is {@code true}, will specify the target version to which
+     * the parent artifact will be updated.</p>
+     * @since 2.17.0
+     */
+    @Parameter(property = "parentVersion")
+    protected String parentVersion = null;
```
slawekjaranowski, line 75: Can be parameter fields `private`?

cachescrubber: My aim was to be as close to `UpdateParentMojo` as possible. All parameters are protected there as well. Should I change them to `private` visibility in both classes?

ajarmoniuk: Guilty as charged. Makes it easier to test indeed without the need to use reflection.

cachescrubber: So let's leave it like this?

slawekjaranowski: ok

slawekjaranowski (approved):

PR #1051, opened 2024-02-17
Resolves #1042: Fixed set logic wrt processAllModules

### LINKED ISSUE #1042: versions:set does behave wrong when parent version is the same as the version to be set
Hi,

I figured out a strange behaviour of the versions plugin (tested with latest release version 2.16.2). 

When I try to set the version in the following way for a project where the current version is 2024.1.0-SNAPSHOT:

`mvn -DgenerateBackupPoms=false '-DnewVersion=2024.1.0' 'org.codehaus.mojo:versions-maven-plugin:2.16.2:set'   -DprocessAllModules '-Dmaven.deploy.skip=true' -DskipITs '-Djacoco.skip=true'`

with maven 3.9.6 and jdk11, nothing is changed. The result looks like this:

```
[INFO] Searching for local aggregator root...
[INFO] Local aggregation root: D:\temp\maven-version\childproject\reactor
[INFO] Processing change of my.example.com.concrete:concrete-reactor:2024.1.0-SNAPSHOT -> 2024.1.0
[INFO] Processing my.example.com.concrete:concrete-reactor
[INFO]     Updating project my.example.com.concrete:concrete-reactor
[INFO]         from version 2024.1.0-SNAPSHOT to 2024.1.0
[INFO]     Updating project my.example.com.concrete:concrete-reactor
[INFO]         from version 2024.1.0 to 2024.1.0-SNAPSHOT
[INFO]
[INFO] Processing my.example.com.concrete:concrete-webapp
[INFO]     Updating project my.example.com.concrete:concrete-webapp
[INFO]         from version 2024.1.0-SNAPSHOT to 2024.1.0
[INFO]     Updating project my.example.com.concrete:concrete-webapp
[INFO]         from version 2024.1.0 to 2024.1.0-SNAPSHOT
```

When I use a different version like this (use version 2024.1.1 instead of 2024.1.0)

`mvn -DgenerateBackupPoms=false '-DnewVersion=2024.1.1' 'org.codehaus.mojo:versions-maven-plugin:2.16.2:set'   -DprocessAllModules '-Dmaven.deploy.skip=true' -DskipITs '-Djacoco.skip=true'`

the output looks like this and it works as expected:

```
[INFO] --- versions:2.16.2:set (default-cli) @ concrete-reactor ---
[INFO] Searching for local aggregator root...
[INFO] Local aggregation root: D:\temp\maven-version\childproject\reactor
[INFO] Processing change of my.example.com.concrete:concrete-reactor:2024.1.0-SNAPSHOT -> 2024.1.1
[INFO] Processing my.example.com.concrete:concrete-reactor
[INFO]     Updating project my.example.com.concrete:concrete-reactor
[INFO]         from version 2024.1.0-SNAPSHOT to 2024.1.1
[INFO]
[INFO] Processing my.example.com.concrete:concrete-webapp
[INFO]     Updating project my.example.com.concrete:concrete-webapp
[INFO]         from version 2024.1.0-SNAPSHOT to 2024.1.1
```

Project looks like this:

```
mainreactor
  |
  |   pom.xml (version 2024.1.0, no parent)
  |
  |---- pom-parent
  |         |
  |         |   pom.xml (version 2024.1.0 no parent)
  |
  |---- package-parent
  |         |
  |         |   pom.xml (version 2024.1.0, pom-parent as parent)
   
 
childreactor
  |
  |   pom.xml (version 2024.1.0-SNAPSHOT, parent mainreactor:pom-parent:2024.1.0)
  |
  |---- child-webapp
  |          |
  |          |   pom.xml (version 2024.1.0-SNAPSHOT, parent mainreactor:package-parent:2024.1.0)

```

I expect when setting the version via `versions:set` to 2024.1.0 the versions in the pom's from childproject are set to 2024.1.0 but they where "reset" to 2024.1.0-SNAPSHOT.

Main background: We using gitflow plugin to create releases and hotfixes and the gitflow plugin itself uses the versions plugin to update the versions to release and snapshot back. Because the versions plugin does not update the pom.xml the gitflow plugin fails because no changes where made.

See minimal example project which shows the behaviour. Please first run mvn install in mainproject/reactor and then go to childproject/reactor and try something like this:

`mvn -DgenerateBackupPoms=false '-DnewVersion=2024.1.0' 'org.codehaus.mojo:versions-maven-plugin:2.16.2:set'   -DprocessAllModules '-Dmaven.deploy.skip=true' -DskipITs '-Djacoco.skip=true'`

[maven-version.zip](https://github.com/mojohaus/versions/files/14095305/maven-version.zip)

jarmoniuk:
#1042 is a pretty particular case which exposes a bug in the set logic.

The new version is initially set "to be tried out". Then, it is being set for the second time when processing changes. However, there's also some logic which is supposed to see if there are modules using the module that is being changed. This logic was responsible for the reversed change in the case of this particular issue. 

When investigating, I've noticed that `processAllModules` has two meanings: one is to process two modules regardless whether they match the current module; the second makes the matching logic traverse the whole reactor instead of the current module's subtree in search of modules which use the current module as parent. Because of that, I have introduced a new option, `fullTreeMatch`, which should by default be `false` as I don't suppose it to be used very often.

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
@@ -25,20 +25,24 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TimeZone;
 import java.util.TreeMap;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.model.Model;
-import org.apache.maven.model.Parent;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Mojo;
@@ -334,6 +338,7 @@ public class SetMojo extends AbstractVersionsUpdaterMojo {
             Map<File, Model> reactorModels = PomHelper.getChildModels(project, getLog());
             final SortedMap<File, Model> reactor = new TreeMap<>(new ReactorDepthComparator(reactorModels));
             reactor.putAll(reactorModels);
+            final Map<Pair<String, String>, Set<Model>> children = computeChildren(reactorModels);
 
             // set of files to update
             final Set<File> files = new LinkedHashSet<>();
@@ -379,8 +384,8 @@ public class SetMojo extends AbstractVersionsUpdaterMojo {
                                 || oldVersionIdRegex.matcher(mVersion).matches())
                         && !newVersion.equals(mVersion)) {
                     applyChange(
-                            project,
                             reactor,
+                            children,
                             files,
                             mGroupId,
                             m.getArtifactId(),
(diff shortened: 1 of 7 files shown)


versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
```diff
@@ -489,7 +497,7 @@ private void applyChange(
             getLog().debug("Looking for modules which use "
                     + ArtifactUtils.versionlessKey(sourceGroupId, sourceArtifactId) + " as their parent");
 
-            for (Map.Entry<File, Model> stringModelEntry : processAllModules
+            for (Map.Entry<File, Model> stringModelEntry : fullTreeMatch
```
slawekjaranowski, line 500: If we have `fullTreeMatch` set to `true` we will iterate again by `reactor.entrySet()` we have loop in loop 😄 
see line: 470

slawekjaranowski:
> fullTreeMatch, which should by default be false as I don't suppose it to be used very often.

maybe not add new parameter ... 
when we add we also should add a test which show what for and what for cases it is .used ..

slawekjaranowski (approved):

PR #1089, opened 2024-06-16
Require Maven 3.6.3, dependencies cleanups

slawekjaranowski:
- follow hints from dependency:analyze


slachiewicz (approved):
huge clen-up - nice work.

PR #1120, opened 2024-07-18
Refactor - Remove redundant `InputStreamReader`

JackPGreen:
Rather than converting `Files.newInputStream` to a `BufferedReader` via an intermediate `InputStreamReader`, `Files.newBufferedReader` can be used instead.

This _currently_ works the same internally in the JRE.


slachiewicz (approved):

JackPGreen:
Integration tests fail, clearly not a like-for-like change.

Issue #1121, opened 2024-07-24
versions:resolve-ranges very slow when parents contain lots of version properties


m-schutte-ohra-nl:
versions:2.17.1:resolve-ranges

The resolve-ranges goal takes a lot of time processing properties that are used in dependency versions (let's call them version properties) in parent projects . As far as I can tell the only version properties that need to be considered are those occurring in the pom of the current project because those are the only ones that could possibly be updated.

When using a parent with a lot of version properties (eg. spring-boot-dependencies) this leads to execution times of several minutes. This has already been reported for other mojos.

For instance the following pom takes 01:23 min and hits the repository with several hundreds of maven-metadata.xml requests.

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <parent>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-dependencies</artifactId>
    <version>3.3.2</version>
  </parent>
  <artifactId>it-resolve-ranges-issues-todo</artifactId>
  <packaging>pom</packaging>

  <properties>
    <htmlunit.version>[2.6,3)</htmlunit.version>
  </properties>
</project>

```
Browsing the history of issues and pull requests I see the introduction of 'includeParents' in several mojos and methods (as a solution to for instance #367) to have caused this.

I don't think introduction of the includeParents parameter to the resolve-ranges mojo is the best solution, because as already stated in #837, setting this to false will sometimes lead to incorrect results. Instead, I am preparing a merge request that restricts the properties to consider to those that exist in the current project.

khmarbaise:
You should configure the resolution of transitive deps etc. via configuration 
https://www.mojohaus.org/versions/versions-maven-plugin/resolve-ranges-mojo.html
And it might be a good idea to ignore the dependencyManagement depending your use case:
```xml
            <configuration>
              <processDependencyManagementTransitive>false</processDependencyManagementTransitive>
            </configuration>
```

PR #1122, opened 2024-07-24
Improve speed of versions:resolve-ranges in case parents contain many version properties

m-schutte-ohra-nl:
PR for #1121 

If 'includeProperties' is null, use it to restrict the properties that are considered in the resolve-ranges mojo to those from the current project. Otherwise, user will normally have included only a subset of those.
Also added an it to cover the situation in #367 as this wasn't covered yet and led me to believe at first that setting 'includeParent' to false in the VersionPropertiesMapRequest was enough.

### DIFF
--- a/versions-maven-plugin/src/it/it-resolve-ranges-issue-1121/verify.bsh
+++ b/versions-maven-plugin/src/it/it-resolve-ranges-issue-1121/verify.bsh
@@ -1,0 +1,21 @@
+import java.io.*;
+import org.codehaus.plexus.util.FileUtils;
+
+try
+{
+    File file = new File( basedir, "pom.xml" );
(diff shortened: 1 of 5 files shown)


slawekjaranowski:
@m-schutte-ohra-nl  thanks for PR

What title of PR should be ... ?

Please consider that it will be used in release notes, so what we want to say for users?

versions-maven-plugin/src/it/it-resolve-ranges-issue-1121/verify.bsh
slawekjaranowski, line 1: I prefer groovy than bsh ... but it is also ok

slawekjaranowski (approved):

PR #1132, opened 2024-08-20
Resolves #1131: Suppress logging when allowMajorUpdates, allowMinorUp…

### LINKED ISSUE #1131: Too aggressive logging of `Assuming allowMajorUpdates false because allowMinorUpdates is false`
I have a project utilize the goal `display-dependency-updates` to check minor and incremental dependency updates. The config has explicitly set `allowMajorUpdates` to false, but on running
the goal the plugin still emit lots of `Assuming allowMajorUpdates false because allowMinorUpdates is false`.

```
[INFO] --- versions:2.17.1:display-dependency-updates (default) @ foo ---
[INFO] Assuming allowMajorUpdates false because allowMinorUpdates is false.
[INFO] Assuming allowMajorUpdates false because allowMinorUpdates is false.
[INFO] Assuming allowMajorUpdates false because allowMinorUpdates is false.
[INFO] Assuming allowMajorUpdates false because allowMinorUpdates is false.
[INFO] Assuming allowMajorUpdates false because allowMinorUpdates is false.
[INFO] Assuming allowMajorUpdates false because allowMinorUpdates is false.
[INFO] Assuming allowMajorUpdates false because allowMinorUpdates is false.
...
```

I think those logging can be omitted if user set `allowMajorUpdates` to false explicitly.

gmshake:
…dates and allowIncrementalUpdates are in valid combination

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/utils/SegmentUtils.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/utils/SegmentUtils.java
@@ -64,11 +64,11 @@ public class SegmentUtils {
      */
     public static Optional<Segment> determineUnchangedSegment(
             boolean allowMajorUpdates, boolean allowMinorUpdates, boolean allowIncrementalUpdates, Log log) {
-        if (log != null && !allowIncrementalUpdates) {
+        if (log != null && !allowIncrementalUpdates && allowMinorUpdates) {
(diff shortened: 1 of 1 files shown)


slawekjaranowski (approved):

gmshake:
There're more places to be **FIXED** by this pattern but I think they should be refactored firstly and all of them should get fixed once.

slawekjaranowski:
@gmshake thanks

Issue #1018, opened 2023-10-26
Regression in 2.16.1: class java.util.HashMap$Node cannot be cast to class java.util.HashMap$TreeNode


RalfNaujokat:
Hi,

i found a regression from 2.16.0 to 2.16.1.

While running display-dependency-updates, i get
Unable to acquire metadata for dependencies java.util.stream.ReferencePipeline$2@35ee466f:
java.lang.ClassCastException: class java.util.HashMap$Node cannot be cast to class java.util.HashMap$TreeNode (java.util.HashMap$Node and java.util.HashMap$TreeNode are in module java.base of loader 'bootstrap')

The last visible action was:
Downloading from ***/io/prometheus/simpleclient/maven-metadata.xml

greetings
Ralf

slawekjaranowski:
Can you provide a simple project which can reproduce it?

RalfNaujokat:
Actually, i can't reproduce it myself, right now it is working.
Could be some cached broken files?
Vers. 2.16.0 did not had the problem, but 2.16.1 crashed.

Sorry for that, maybe we can close this issue,
and if i notice this again, i will keep you informed.

slawekjaranowski:
ok, if you can reproduce please reopen.

jgraglia:
Hi,
I too have the same exception with one of my jenkins job, while generating a mvn site with version reports.
I struggle to replicate this problem on my box and the jenkins env is a bit complex (docker cloud..) to replicate
I managed to get the following trace by activating the error reporting in maven (`mvn --errors...`)  : 
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:4.0.0-M9:site (default-site) on project xxxxxxxx: Error generating versions-maven-plugin:2.16.1:dependency-updates-report report: org.codehaus.mojo.versions.api.VersionRetrievalException: Unable to acquire metadata for dependencies java.util.stream.ReferencePipeline$2@405e3e35: java.lang.ClassCastException: java.util.HashMap$Node cannot be cast to java.util.HashMap$TreeNode -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-site-plugin:4.0.0-M9:site (default-site) on project xxxxxxxx: Error generating versions-maven-plugin:2.16.1:dependency-updates-report report
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
[...]

PR #1133, opened 2024-08-20
bugix/ protect internal data in multi thread env.

jgraglia:
Map is not thread safe but some mojo are marked thread safe.

This PR use a safe structure.

_PS: null check added because ConcurrentHashMap does not accept null value (diff with HashMap)_


jgraglia:
This PR is expected to fix https://github.com/mojohaus/versions/issues/1018#issuecomment-2248374042

PR #1187, opened 2024-11-23
Fixes #1140: Added showVersionless (default true)

### LINKED ISSUE #1140: Enhancement: limit report to updates actionable in the context of pom.xml
I found nothing similar in closed issues, if you consider this a duplicate to a previous issue, feel free to close it.

**Motivation:**
I'm trying to limit the reports down to a point that all listed dependencies, plugins, parent and properties version are specified in the analyzed pom.xml. Basically all inherited (upstream) issues should be ignored. I want to make it as simple and fast as possible to fix such issues without the need to reason about if a reported issue is inherited from upstream.

**Example**: two projects, from different reactors

* parent/pom.xml uses spring-boot-starter-parent:3.3.1, but 3.3.3 is latest version
* service/pom.xml uses parent/pom.xml at latest version

report for
* parent/pom.xml 
  * MUST list outdated parent
  * MUST NOT list e.g. outdated plugins, dependencies that are inherited from spring-boot-starter-parent
* service/pom.xml
  * MUST NOT list e.g. outdated parent, plugins etc. inherited from parent/pom.xml

**Possible Solutions**:

* preferred, additional flag e.g. _ignoreInherited_ to control this behavior
* alternative, mark "inherited" issues in reports (and possibly xml) as such

andrzejj0:
Resolves #1140. New optional parameter, `showVersionless` default `true` so that we keep backward compatibility.

Scope of `verbose` is enhanced: with `showVersionless` or with `processDependencyManagementTransitive`, it will also show where the version is managed (groupId:artifactId:version of the POM which manages the version of the dependency).

Please review.

I'm also working on another branch which will use a common service (for dependency updates, dependency updates reports, and possibly for updating those versions as well) to retrieve dependency updates. This will bring cohesion much needed cohesion.


andrzejj0:
Converting to draft: will add this to report as well.

pom.xml
```diff
@@ -10,7 +10,7 @@
 
   <groupId>org.codehaus.mojo.versions</groupId>
   <artifactId>versions</artifactId>
-  <version>2.18.0</version>
+  <version>2.18.1-SNAPSHOT</version>
```
slawekjaranowski, line 13: should be in code .... why such change
please rebase with current master

andrzejj0: probably a bad rebase

andrzejj0: rebased

slawekjaranowski (approved):

PR #1188, opened 2024-11-24
Fixes #1140 for reports

### LINKED ISSUE #1140: Enhancement: limit report to updates actionable in the context of pom.xml

andrzejj0:
I managed to reproduce #1140 for reports, so here is a fix.

Added an integration test + showVersionless for dependency-updates-report.

Added some small refactoring.

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractDependencyUpdatesReport.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractDependencyUpdatesReport.java
@@ -43,6 +43,7 @@ import org.codehaus.mojo.versions.api.VersionRetrievalException;
 import org.codehaus.mojo.versions.reporting.ReportRendererFactory;
 import org.codehaus.mojo.versions.reporting.model.DependencyUpdatesModel;
 import org.codehaus.mojo.versions.utils.DependencyComparator;
+import org.codehaus.mojo.versions.utils.MavenProjectUtils;
 import org.codehaus.mojo.versions.xml.DependencyUpdatesXmlReportRenderer;
 import org.codehaus.plexus.i18n.I18N;
 import org.eclipse.aether.RepositorySystem;
@@ -79,6 +80,16 @@ public abstract class AbstractDependencyUpdatesReport extends AbstractVersionsRe
      */
     @Parameter(property = "processDependencyManagementTransitive", defaultValue = "true")
     protected boolean processDependencyManagementTransitive;
+
+    /**
+     * <p>Include dependencies with version set in a parent or in a BOM.</p>
+     * <p>This is similar to {@code processDependencyManagementTransitive}, but will
+     * report updates on dependencies.</p>
+     *
+     * @since 2.19.0
+     */
+    @Parameter(property = "showVersionless", defaultValue = "true")
+    protected boolean showVersionless = true;
 
     /**
      * Report formats (html and/or xml). HTML by default.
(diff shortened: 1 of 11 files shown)


versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractDependencyUpdatesReport.java
```diff
@@ -80,6 +81,17 @@ public abstract class AbstractDependencyUpdatesReport extends AbstractVersionsRe
     @Parameter(property = "processDependencyManagementTransitive", defaultValue = "true")
     protected boolean processDependencyManagementTransitive;
 
+    /**
+     * <p>Include dependencies with version set in a parent or in a BOM.</p>
+     * <p>This is similar to {@code processDependencyManagementTransitive}, but will
+     * report updates on dependencies.</p>
+     * <p>Default is {@code false}.</p>
```
slawekjaranowski, line 88: When we use `defaultValue` in `@Parameter` is will be added to docs.
By the way has different value .... doc and parameter

andrzejj0: I get your point, but I did not change `processDependencyManagementTransitive` in this PR.

On the other hand, this (initializing the properties) immensely helps with unit testing.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractDependencyUpdatesReport.java
```diff
@@ -80,6 +81,17 @@ public abstract class AbstractDependencyUpdatesReport extends AbstractVersionsRe
     @Parameter(property = "processDependencyManagementTransitive", defaultValue = "true")
     protected boolean processDependencyManagementTransitive;
 
+    /**
+     * <p>Include dependencies with version set in a parent or in a BOM.</p>
+     * <p>This is similar to {@code processDependencyManagementTransitive}, but will
+     * report updates on dependencies.</p>
+     * <p>Default is {@code true}.</p>
```
slawekjaranowski, line 88: it is not needed will be generated .... check site output 😄

andrzejj0: Removed <p>Default is {@code true}.</p>

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractDependencyUpdatesReport.java
```diff
@@ -80,6 +81,16 @@ public abstract class AbstractDependencyUpdatesReport extends AbstractVersionsRe
     @Parameter(property = "processDependencyManagementTransitive", defaultValue = "true")
     protected boolean processDependencyManagementTransitive;
 
+    /**
+     * <p>Include dependencies with version set in a parent or in a BOM.</p>
+     * <p>This is similar to {@code processDependencyManagementTransitive}, but will
+     * report updates on dependencies.</p>
+     *
+     * @since 2.18.1
```
slawekjaranowski, line 89: Maybe should be a 2.19.0 as we add new feature ... what do you think?

slawekjaranowski: or it is for fixing bug only ...

andrzejj0: Ok. Changed that in DisplayDependencyUpdatesMojo as well.

andrzejj0: Well, a bug fix and a feature toggle. I don't know :-)

PR #1193, opened 2024-11-27
Fixed #1191: Processes (all) profiles from original model if transitive is false

### LINKED ISSUE #1191: `display-dependency-updates` with `processDependencyManagementTransitive = false` ignores `dependencyManagement` in enabled profiles
When running `display-dependency-updates` with `processDependencyManagementTransitive` is disabled (i.e. `false`), `dependencyManagement` in enabled profiles is not checked for updates.

andrzejj0:
If processDependencyManagementTransitive is false, will also process dependency management entries from profiles (all profiles). 
This was previously reported as a bug.

### DIFF
--- a/versions-maven-plugin/src/it/it-dependency-updates-report-issue-1191/verify.groovy
+++ b/versions-maven-plugin/src/it/it-dependency-updates-report-issue-1191/verify.groovy
@@ -1,0 +1,8 @@
+import groovy.xml.XmlSlurper
+
+def report = new XmlSlurper().parse(new File(basedir, "target/dependency-updates-report.xml"))
+assert !(report
+        .dependencyManagements
+        .dependencyManagement
+        .find { node -> node.artifactId == 'dummy-api' }.isEmpty())
+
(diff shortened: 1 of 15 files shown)


versions-maven-plugin/src/it/it-dependency-updates-report-issue-1191/verify.groovy
```diff
@@ -0,0 +1,3 @@
+assert new File( basedir, "target/dependency-updates-report.xml" ).text
+        .replaceAll( '\\s*', '' )
+        .contains( '<artifactId>dummy-api</artifactId>' )
```
slawekjaranowski, line 3: for xml we can use `XmlSlurper`

andrzejj0: Ok, so while doing this I discovered that I was using XmlSlurper's find wrong. So corrected that too.

slawekjaranowski (approved):

PR #1194, opened 2024-11-30
Fixed the failed release build

andrzejj0:
Several issues with integration tests for #333 caused it to break the release build with Maven 4.


versions-maven-plugin/src/it/it-dependency-updates-aggregate-report-issue-333/invoker.properties
```diff
@@ -1 +1 @@
-invoker.goals=site
+invoker.goals = ${project.groupId}:${project.artifactId}:${project.version}:dependency-updates-aggregate-report
```
slawekjaranowski, line 1: Does it will work with site plugin?
Or only as direct report call?

andrzejj0: Yes, it does. The only difference is the location of the generated files (site vs reports).

PR #1202, opened 2024-12-24
Refactoring: coalesce use-*-* goals.

andrzejj0:
I wanted to try and do something similar to what mojohaus/versions#292 was, but since the plugin has moved on, it first needs some refactoring.

This PR makes the use-\*-\* goals use shared codebase in order to make future refactoring simpler.

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojo.java
@@ -16,37 +16,20 @@ package org.codehaus.mojo.versions;
  */
 
 import javax.inject.Inject;
-import javax.xml.stream.XMLStreamException;
 
-import java.io.IOException;
-import java.util.Collection;
+import java.util.Comparator;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
+import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.model.Dependency;
-import org.apache.maven.model.DependencyManagement;
-import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.plugin.MojoFailureException;
-import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.wagon.Wagon;
-import org.codehaus.mojo.versions.api.PomHelper;
-import org.codehaus.mojo.versions.api.Segment;
-import org.codehaus.mojo.versions.api.VersionRetrievalException;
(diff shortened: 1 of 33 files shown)


versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojo.java
```diff
@@ -62,7 +45,7 @@ public class UseLatestVersionsMojo extends UseLatestVersionsMojoBase {
      * @since 1.2
      */
     @Parameter(property = "allowMajorUpdates", defaultValue = "true")
-    private boolean allowMajorUpdates = true;
+    protected boolean allowMajorUpdates = true;
```
slawekjaranowski, line 48: if we introduce also geters for it I would like to left it private

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseNextSnapshotsMojo.java
```diff
@@ -98,76 +80,43 @@ public UseNextSnapshotsMojo(
...
+    }
+
+    @Override
+    protected boolean isAllowIncrementalUpdates() {
+        return allowIncrementalUpdates;
+    }
+
+    @Override
+    protected boolean isAllowSnapshots() {
+        return true;
```
slawekjaranowski, line 100: why such ...

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojo.java
andrzejj0: I actually exposed it for the tests, so that we wouldn't need to use reflection to manipulate those.

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseNextSnapshotsMojo.java
andrzejj0: This is a callback for [UseLatestVersionsMojoBase#update](https://github.com/mojohaus/versions/pull/1202/files#diff-67c5e18947ef94a25779821e3341fd0f5660aa4031c8c1492b252c09cf0f0716R153) so that it selects snapshots -- we are looking for snapshots to upgrade to.

andrzejj0: The property `allowSnapshots` doesn't make sense in some of the goals, like this one, so I'll push the property down the class hierarchy to goals where it does make sense. At the same time, I'll pull `isAllowSnapshots()` upwards as an abstract method.

andrzejj0: Please note that I'm doing this in stages so that the PR doesn't grow too big.

andrzejj0: test comment

slawekjaranowski (approved):

PR #1204, opened 2024-12-25
Versions refactor 2

andrzejj0:
This is a second stage of the other PR, adding UpdateParent to the common base: I've noticed that updating parent is already being done by Use-\*-\*, but it used a different method. So why don't we refactor it even further? The goal is to make the plugin codebase smaller.

Also, here I have replaced `is...` getters by `get...` for booleans as they were not always grammatical and not always consistent (`isProcessing...` which differs from the name of the field).

Edit: added a comment: with skipResultion == false, this goal is a slightly modified use-\*-\* targeting parent only, with an added modification that it also has the `parentVersion` parameter.

Maybe I should just merge this PR into the other one and create one big PR instead?


slawekjaranowski:
can be done step by step.
Please merge depends PR and resolve conflicts.

andrzejj0:
Rebased.

One remark: because the goal now shares the common base with other similar goals, which don't print such information, we are now also not logging INFO lines like:

`[INFO] The parent project is the latest version`

so I had to remove a check for that line from one of the it tests.

slawekjaranowski (approved):

PR #1207, opened 2025-01-01
Parallel version updates

andrzejj0:
This was actually suggested earlier for property updates.

Update retrieval (which is immutable) is split from update execution which made parallel updates feasible.

`ConcurrentLinkedQueue` is chosen as the structure for appending updates (it's being written to by multiple threads).

Introduced a new parameter, `numThreads`.

Also, bumped version to `2.19.0-SNAPSHOT`

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojoBase.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojoBase.java
@@ -18,10 +18,15 @@ package org.codehaus.mojo.versions;
 import javax.xml.stream.XMLStreamException;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojoBase.java
```diff
@@ -74,6 +89,64 @@ public UseLatestVersionsMojoBase(
         super(artifactHandlerManager, repositorySystem, wagonMap, changeRecorders);
     }
 
+    /**
+     * Represents a version change of an artifact
+     */
+    public abstract static class ArtifactVersionChange {
```
slawekjaranowski, line 95: Should such class be public ... looks like can be use as private

andrzejj0: Done

slawekjaranowski (approved):

PR #1212, opened 2025-01-28
Removal of Mercury and Numeric version schemes, refactoring

andrzejj0:
Removed Numeric and Mercury version schemes. Mercury is probably not used at all, Numeric can be used due to some artifacts, but since it is not respected by Maven Resolver itself, it makes little sense to keep it.

With the two comparators removed, we can simply use the default comparator that `DefaultArtifact` provides on its own (which is also in line with Resolver implementation of `GenericVersion`).

The rest is refactoring:
- Simplified code, added some unit tests while doing so;
- Extracted `RuleService` from `VersionsHelper` to handle rules-related business, e.g. ignored versions;
- Extracted `ArtifactFactory`;
- Renamed `DefaultArtifactVersionCache` to `ArtifactVersionService`;
- `PropertyVersions`: moved artifact resolution business to `DefaultVersionsHelper`;


slawekjaranowski:
great, as it is next big change I will try to review it in next weekend
of course will be good if someone else also look at it

andrzejj0:
Please feel free to invite other reviewers. Maybe Hervé or Tamas?

versions-common/src/main/java/org/codehaus/mojo/versions/api/ArtifactVersionsCache.java
```diff
@@ -16,7 +16,7 @@
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
- * under the License.
+ * under the License.T
```
slawekjaranowski, line 19: 😄

andrzejj0:
> can you split for smaller one ... eg, looks like can be done separatly
> 
>     * Extracted ArtifactFactory;
> 
>     * Renamed DefaultArtifactVersionCache to ArtifactVersionService;

Created a PR including only those two, but it's still quite sizeable mostly due to the fact that we're refactoring a key part of this whole plugin -- even though the change is small.

PR #1221, opened 2025-03-08
Pom helper refactoring

andrzejj0:
Second stage of refactoring, this time mainly focused on PomHelper


versions-common/src/main/java/org/codehaus/mojo/versions/api/PropertyVersions.java
```diff
@@ -59,61 +56,33 @@ public class PropertyVersions extends AbstractVersionDetails {
...
+    PropertyVersions(
+            String profileId,
+            String name,
+            Log log,
+            VersionsHelper helper,
+            Set<ArtifactAssociation> associations,
+            SortedSet<ArtifactVersion> resolvedVersions) {
         this.profileId = profileId;
         this.name = name;
+        this.log = log;
```
slawekjaranowski, line 81: should we check for null value as is in other places ...

versions-common/src/main/java/org/codehaus/mojo/versions/api/PropertyVersionsBuilder.java
```diff
@@ -97,120 +103,37 @@ public String getName() {
...
 
     public PropertyVersionsBuilder withLowerBound(String lowerBound, boolean includeLower) {
-        Boolean value = lowerBounds.get(lowerBound);
-        if (value == null) {
-            value = includeLower;
-        } else {
-            value = includeLower && value;
-        }
-        lowerBounds.put(lowerBound, value);
+        lowerBounds.compute(lowerBound, (__, oldValue) -> ofNullable(oldValue).orElse(true) && includeLower);
```
slawekjaranowski, line 128: `__` it is a correct variable name 😄 I still learning ....

slawekjaranowski (approved):

versions-common/src/main/java/org/codehaus/mojo/versions/api/PropertyVersions.java
andrzejj0: That's for later. Doing so would make us have to provide this in all the tests and I still wanted to minimize the size of this PR.

Issue #1147, opened 2024-09-23
mvn  versions:display-dependency-updates throws llegalArgumentException: Comparison method violates its general contract!


RalfNaujokat:
Using Maven 3.9.9 with different JDKs (Oracle Java 21, GraalVM 21/23) on Linux

StackTrace is:

`
== BEGIN ==

[ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.17.1:display-dependency-updates (default-cli) on project libraries: Unable to acquire metadata for dependencies java.util.stream.ReferencePipeline$2@629591d0: java.lang.IllegalArgumentException: Comparison method violates its general contract! -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.17.1:display-dependency-updates (default-cli) on project libraries: Unable to acquire metadata for dependencies java.util.stream.ReferencePipeline$2@629591d0: java.lang.IllegalArgumentException: Comparison method violates its general contract!
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:193)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:180)
[...]

PR #1223, opened 2025-03-15
Site fix for rule-2.1.0.xsd and rule-2.0.0.xsd; getOutputPath() for reports

andrzejj0:
Added local xsd files for rule 2.1.0 and 2.0.0.
Also, using getBundleName instead of deprecated getOutputName for reports


versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/AbstractVersionsReport.java
```diff
@@ -209,6 +209,12 @@ protected void executeReport(Locale locale) throws MavenReportException {
         }
     }
 
+    /**
+     * Returns the bundle name for the underlying report renderers
+     * @return bundle name for the underlying report renderers
+     */
+    protected abstract String getBundleName();
+
```
slawekjaranowski, line 217: in `org.apache.maven.reporting.MavenReport` there is replaced `getOutputName` by `getOutputPath` - maybe we also use `getOutputPath` here ...

andrzejj0: I don't remember the reason why I did that. I'll see if we can use `getOutputPath`.

slawekjaranowski (approved):

slawekjaranowski:
@andrzejj0  - Is it time for next release? I think so...

andrzejj0:
> @andrzejj0 - Is it time for next release? I think so...

Almost: I'm working on a low hanging fruit - adding the information on the artifact causing problems for the resolver for #1147 -- just writing unit tests testing it at the moment.

PR #1224, opened 2025-03-16
Exposing the problem causing artifact if Resolver throws an exception

andrzejj0:
Any problems causing the Resolver to fail resolving a version should be reported to the end user,  including the artifact causing the problem which can facilitate problem resolution on the resolver end. 

Here we're propagating the information about the problematic artifact to the mojo which can then output it in the log.

Related: #1147

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/VersionRetrievalException.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/VersionRetrievalException.java
@@ -19,17 +19,26 @@ package org.codehaus.mojo.versions.api;
  * under the License.
  */
 
+import java.util.Optional;
+
+import org.apache.maven.artifact.Artifact;
+
 /**
  * Exception thrown if version information cannot be retrieved
  */
 public class VersionRetrievalException extends Exception {
+
+    private final Artifact artifact;
+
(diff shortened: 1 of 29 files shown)


andrzejj0:
@slawekjaranowski please check; we can release after this one is merged.

versions-common/src/main/java/org/codehaus/mojo/versions/api/VersionRetrievalException.java
```diff
@@ -69,12 +81,30 @@ public VersionRetrievalException(Throwable cause) {
...
+     */
+    public Optional<Artifact> getArtifact() {
+        return Optional.ofNullable(artifact);
+    }
+
+    /**
+     * Sets the artifact causing the problem with version retrieval
+     * @param artifact {@link Artifact} instance causing the problem
+     */
+    public void setArtifact(Artifact artifact) {
```
slawekjaranowski, line 107: exception should be immutable, so I would like to declare artifact as final and use only constructors ...

andrzejj0: Done

slawekjaranowski (approved):

PR #1234, opened 2025-04-11
Fixed #1233- set-property in profile without module  property

### LINKED ISSUE #1233: set-property doesn't work as expected if the property is not present in the module
Considering the pom:

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <groupId>default-group</groupId>
    <artifactId>default-artifact</artifactId>
    <version>1.0</version>
    <packaging>pom</packaging>

    <profiles>
        <profile>
            <id>test-profile</id>
            <properties>
                <dummy-api-version>test-value</dummy-api-version>
            </properties>
            <dependencies>
                <dependency>
                    <groupId>localhost</groupId>
                    <artifactId>dummy-api</artifactId>
                    <version>${dummy-api-version}</version>
                </dependency>
            </dependencies>
        </profile>
    </profiles>
</project>

```
running

`mvn versions:set-property -Dproperty=dummy-api-version -DnewVersion=1.2.3 -DgenerateBackupPoms=false -DprofileId=test-profile`

I'm expecting that the `dummy-api-version` property will be `1.2.3` but it doesn't

to make this works it is necessary to add the same property on the module

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <groupId>default-group</groupId>
    <artifactId>default-artifact</artifactId>
    <version>1.0</version>
    <packaging>pom</packaging>
    <properties>
        <dummy-api-version>test-value</dummy-api-version>
    </properties>

    <profiles>
        <profile>
            <id>test-profile</id>
            <properties>
                <dummy-api-version>test-value</dummy-api-version>
            </properties>
            <dependencies>
                <dependency>
                    <groupId>localhost</groupId>
                    <artifactId>dummy-api</artifactId>
                    <version>${dummy-api-version}</version>
                </dependency>
            </dependencies>
        </profile>
    </profiles>
</project>
```

even if it changes the value in the profile, it should works even if the property is declared only in the profile

mcarlett:
Handles the case when the property is not in the project properties but it is present in the profile properties

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
@@ -171,8 +171,19 @@ public class SetPropertyMojo extends AbstractVersionsUpdaterMojo {
             Property currentProperty = entry.getKey();
             PropertyVersions version = entry.getValue();
             String newVersionGiven = currentProperty.getVersion();
-            final String profileToApply = isEmpty(profileId) ? version.getProfileId() : profileId;
-            final String currentVersion = getProject().getProperties().getProperty(currentProperty.getName());
+            String profileToApply;
+            String currentVersion;
+            if (isEmpty(profileId)) {
+                profileToApply = version.getProfileId();
+                currentVersion = getProject().getProperties().getProperty(currentProperty.getName());
+            } else {
+                profileToApply = profileId;
+                currentVersion = getProject().getModel().getProfiles().stream()
+                        .filter(profile -> profileId.equals(profile.getId()))
+                        .findFirst()
+                        .map(profile -> profile.getProperties().getProperty(currentProperty.getName()))
+                        .orElse(null);
+            }
             if (currentVersion == null) {
                 continue;
             }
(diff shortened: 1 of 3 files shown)


versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java
```diff
@@ -172,7 +173,14 @@ private void update(MutableXMLStreamReader pom, Property[] propertiesConfig, Str
             PropertyVersions version = entry.getValue();
             String newVersionGiven = currentProperty.getVersion();
             final String profileToApply = isEmpty(profileId) ? version.getProfileId() : profileId;
-            final String currentVersion = getProject().getProperties().getProperty(currentProperty.getName());
+            String currentVersion = getProject().getProperties().getProperty(currentProperty.getName());
+            if (currentVersion == null && isNotEmpty(profileId)) {
```
slawekjaranowski, line 177: I will try here with `profileToApply` variable.

mcarlett: I refactored the condition to make the code more readable keeping the condition separated with just one `if` statement, thanks @slawekjaranowski

andrzejj0 (approved):

andrzejj0:
Looks good to me as well. We can squash and merge.

slawekjaranowski (approved):

PR #1264, opened 2025-09-01
Remove manual default from docs for parameters

slawekjaranowski:
It is generated by plugin-tools


andrzejj0 (approved):

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/DisplayDependencyUpdatesMojo.java
```diff
@@ -93,7 +93,6 @@ public class DisplayDependencyUpdatesMojo extends AbstractVersionsDisplayMojo {
      * <p>Include dependencies with version set in a parent or in a BOM.</p>
      * <p>This is similar to {@code processDependencyManagementTransitive}, but will
      * report updates on dependencies.</p>
-     * <p>Default is {@code false}.</p>
```
slawekjaranowski, line 96: this one was different than is set in parameter

Issue #979, opened 2023-06-28
Multiple executions of display-X-updates overwrite outputFile


ctubbsii:
Executing multiple goals in a single command-line causes the most recent one to clobber the selected output file instead of append to it, or add a unique identifier name.

Example:

```bash
mvn clean verify versions:display-parent-updates versions:display-property-updates versions:display-plugin-updates -Dversions.outputFile=target/versions-output.txt
```

In order to get the output for all executions, you must run multiple executions of the plugin with different config, which you can either do with separate command-line executions, or you have to specify different executions in the POM file.

Also, the contents of the file wrap according to the outputLineWidth, when it should not wrap at all when written to a file.

It would be good to have an option for an output file format that is computer-readable, not subject to line width wrapping issues, and such that subsequent executions will append/modify the file, rather than clobber it (they would only clobber the section applicable to that taskk... whether it's properties, plugins, parent, or dependencies).

slawekjaranowski:
@ctubbsii  Thanks for report, I see many issues here:

- overwriting `outputFile`
- wrap according to the o`utputLineWidth` in output files
- output report in computer-readable format

ctubbsii:
@slawekjaranowski That's a good summary. Having a computer readable output that always appends and never wraps would satisfy all three. Something with a comma-delimited format like the following might work:

```
versionPluginGoal,groupId:artifactId:classifier:type,currentVersion,newVersion
```

Example:

```
parent,org.apache:apache:pom,20,29
property,org.apache.maven.plugins:maven-surefire-plugin:jar,2.20.0,3.1.0
property,org.apache.maven.plugins:maven-failsafe-plugin:jar,2.20.0,3.1.0
plugin,org.apache.maven.plugins:maven-enforcer-plugin:jar,1.2.0,3.0.0
dependency,commons-codec:commons-codec:jar,1.0,2.5
```

Could also use XML or JSON as the output format, but that'd be harder to parse with simple command-line tools.

ctubbsii:
This is still relevant.

ctubbsii:
This is still relevant.

andrzejj0:
Outputting to a computer-readable format seems like something in line of a `changeRecorder`, which is currently present in mutating goals. Then, it could be easily extended to provide csv and maybe other formats like json or yaml, so I think I would do that in a separate feature.

andrzejj0:
[...]

PR #1279, opened 2025-09-27
#979: Output file is not overwritten by default

andrzejj0:
Part of #979 restricted to overwriting of the output file.


andrzejj0:
> versions-maven-plugin/src/it/it-display-dependency-updates-009-processDependencyManagementTransitive-true/verify.groovy - why do you remove such assertions?

It looks like I forgot to remove it earlier, in #1187: the test used exactly the same pom.xml as `it-display-dependency-updates-009`, but used the outputFile instead of checking the build log. 

invoker.goals.2=${project.groupId}:${project.artifactId}:${project.version}:display-dependency-updates
invoker.mavenOpts.2=-DprocessDependencyManagementTransitive=true -Dversions.outputFile=./output.2.txt -DoutputEncoding=UTF-8

```
assert output[1] =~ /localhost:dummy-api \.* 1.1 -> 3.0/
assert output[1] =~ /localhost:dummy-impl \.* 1.2 -> 2.2/
assert !(output[1] =~ /localhost:dummy-api-impl-bom-pom \.* 1.0 -> 2.0/)
```
and:

```
invoker.goals=${project.groupId}:${project.artifactId}:${project.version}:display-dependency-updates
invoker.mavenOpts=-DprocessDependencyManagementTransitive=true
```

```
def buildLog = new File( basedir, "build.log").text
assert buildLog =~ /\Qlocalhost:dummy-api\E\s*\.*\s*1\.1\s+->\s+3\.0/
assert buildLog =~ /\Qlocalhost:dummy-impl\E\s*\.*\s*1\.2\s+->\s+2\.2/
assert !(buildLog =~ /\Qlocalhost:dummy-api-impl-bom-pom\E\s*\.*\s*1\.0\s+->\s+2\.0/)
```

Do you think I should do that in a separate PR, or should I restore the redundant it?

andrzejj0:
I've restored it (along with the pom.xml and invoker.properties which were missing); I'll make a small PR that removes that test.

slawekjaranowski (approved):

PR #1283, opened 2025-10-03
Allow filtering out pre releases in use-latest-versions

### LINKED ISSUE #1225: Ignore milestone and pre-releases in use-latest-releases
Some projects publish pre-releases and milestone releases to Maven central. Spring recently started doing this: https://spring.io/blog/2025/01/21/milestones-to-central

Now it means that when you have
```xml
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>org.springframework</groupId>
                <artifactId>spring-framework-bom</artifactId>
                <version>6.2.3</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>
```

then `use-latest-releases` will move you to 7.0.0-M3
```
[INFO] --- versions:2.18.0:use-latest-releases (default-cli) @ project-base ---
[INFO] Updated org.springframework:spring-framework-bom:pom:6.2.3 to version 7.0.0-M3
```

While this might be technically how `use-latest-releases` is specified: "versions that are not snapshots and do not have a year-month-day suffix", it is not exactly what you would want.

It would be great to have an `allowPreReleases` option (default false) similarly to `allowSnapshots` that would ignore versions like `7.0.0-M3` and also at least `1.2.3-alpha12`, `1.2.3-beta1` and `1.2.3-rc`. Then it would work as expected and upgrade to 6.2.4

Artur-:
Adds a new allowPreReleases parameter (defaulting to false) to the use-latest-versions goal to prevent upgrading to pre-release versions (alpha, beta, milestone, rc).

Fixes #1225

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojo.java
@@ -28,6 +28,7 @@ import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.wagon.Wagon;
+import org.codehaus.mojo.versions.api.AbstractVersionDetails;
 import org.codehaus.mojo.versions.api.recording.ChangeRecorder;
 import org.codehaus.mojo.versions.utils.ArtifactFactory;
 import org.eclipse.aether.RepositorySystem;
@@ -84,6 +85,13 @@ public class UseLatestVersionsMojo extends UseLatestVersionsMojoBase {
      */
     @Parameter(property = "allowSnapshots", defaultValue = "false")
     protected boolean allowSnapshots;
+
+    /**
+     * Whether to allow pre-release versions (alpha, beta, milestone, rc) when searching
+     * for the latest version of an artifact.
+     */
+    @Parameter(property = "allowPreReleases", defaultValue = "false")
+    protected boolean allowPreReleases = false;
 
     /**
      * Whether to process the dependencies section of the project.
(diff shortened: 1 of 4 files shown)


versions-maven-plugin/src/test/java/org/codehaus/mojo/versions/UseLatestVersionsMojoTest.java
```diff
@@ -237,4 +237,57 @@ public void testPoisonDependencyVersion()
...
+                        .withGroupId("default-group")
+                        .withArtifactId("dependency-artifact")
+                        .withVersion("0.9.0")
+                        .withType("pom")
+                        .withClassifier("default")
+                        .withScope(SCOPE_COMPILE)
+                        .build()));
+
+        setVariableValueToObject(mojo, "allowSnapshots", true);
+        setVariableValueToObject(mojo, "allowPreReleases", true);
```
slawekjaranowski, line 283: I would like to check only with `allowPreReleases` true to see that beta version will be selected

versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/UseLatestVersionsMojo.java
```diff
@@ -85,6 +86,13 @@ public class UseLatestVersionsMojo extends UseLatestVersionsMojoBase {
     @Parameter(property = "allowSnapshots", defaultValue = "false")
     protected boolean allowSnapshots;
 
+    /**
+     * Whether to allow pre-release versions (alpha, beta, milestone, rc) when searching
+     * for the latest version of an artifact.
+     */
+    @Parameter(property = "allowPreReleases", defaultValue = "false")
```
slawekjaranowski, line 93: There is a new parameter ....
What `since` tag should have?

@andrzejj0

andrzejj0: @since 2.20.0 

(feature release)

andrzejj0 (approved):

PR #1284, opened 2025-10-04
#979: overhaul of the updates logger

andrzejj0:
- **Extended the API for the updates logger**
- New, appendable xml schema which allows recording multiple executions of the plugin
- Added json and csv

The change logger updates are now generated using an XSD schema rather than using Modello. The primary reason was the ability to easily (de)serialize using multiple formats + extensibility.

The ultimate goal of this is to use this change logger model to actually transmit updates in "display" targets as well, so that both "display" and "change" targets share the common backend.

I know it's going to be a large one, so I'm doing this in stages. First stage is this updates overhaul.

@slawekjaranowski  please review


versions-model/src/site/markdown/xsd/index.md
```diff
@@ -11,3 +11,9 @@ Versions - XML schemas for ruleSet
 - [rule-2.1.0.xsd](rule-2.1.0.xsd)
 - [rule-2.0.0.xsd](rule-2.0.0.xsd)
 
+Change recorder updates
+=======================
+
+- [updates-3.0.xsd](updates-3.0.xsd)
+- [updates-2.0.xsd](updates-2.0.xsd)
```
slawekjaranowski, line 18: do we need both version ... does 2.0 will be deprecated?

andrzejj0: Yes, I think updates-2.0.xsd should be deprecated. We currently can generate both and updates-2.0.xsd is still default.

andrzejj0: Updated -- didn't squash.

andrzejj0:
Will add an example of change recorder logs to DisplayDependencyUpdates.

slachiewicz:
Maybe we can use jakarta instead of javax? If it's not a big deal now?

versions-api/pom.xml
```diff
@@ -12,6 +12,19 @@
   <name>Versions API</name>
   <description>Extension API for Versions Plugin</description>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.codehaus.mojo.versions</groupId>
+      <artifactId>versions-model</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
```
slachiewicz: Can we try to avoid bringing this big dependency?

andrzejj0:
> Maybe we can use jakarta instead of javax? If it's not a big deal now?

I tried, but I think that is not possible: we still support Java 8. https://github.com/andrzejj0/versions-maven-plugin/actions/runs/18996348155/job/54256770147

javax.xml.bind:jaxb-api:2.3.1 is the last version that supports Java 8, but it is only compatible with javax.

andrzejj0:
However, it's easy to do it when we eventually drop Java 8 support (some day). 
Meanwhile, added you to reviewers. 

Your Copilot PR would still need to use this new change recorder to record changes, but that's pretty straightforward (a prompt away).

PR #1293, opened 2025-10-15
Added remaining javadoc comments.

andrzejj0:
Also:
- Supressing javadoc linter problems during report generation.
~- Additional javadoc linting step where these are suppressed (it's not possible to have two executions in reporting)~

Currently I'm also enforcing Javadoc correctness, but I'm thinking whether it's not better to just have a lenient linter everywhere.

@slawekjaranowski please approve/advise.


andrzejj0:
Looks like that I can't make javadoc plugin actually see already compiled dependencies; even rearranging modules in parent won't help. I'll remove the additional linter step in favor of a slightly less lenient checks.

slachiewicz (approved):

PR #1327, opened 2025-12-15
Add artifact classifier to ChangeRocorderXML

ggkochanski:
When two dependencies differ by the classifier only then they appear twice in file `version-changes.xml` but you cannot distinguish which is which.

The general idea is to add the artifact's classifier attribute to node `/updates/dependencyUpdate` when is set:

```
        DependencyVersionChange change = (DependencyVersionChange) changeRecord.getVersionChange();
        update.setAttribute("groupId", change.getGroupId());
        update.setAttribute("artifactId", change.getArtifactId());
        if(change.getClassifier() != null) {
            update.setAttribute("classifier", change.getClassifier());
        }

```

### DIFF
--- a/versions-api/src/main/java/org/codehaus/mojo/versions/api/change/DependencyVersionChange.java
+++ b/versions-api/src/main/java/org/codehaus/mojo/versions/api/change/DependencyVersionChange.java
@@ -39,6 +39,13 @@ public interface DependencyVersionChange extends VersionChange {
     String getArtifactId();
 
     /**
+     * Returns the classifier of the dependency
+     * @return classifier of the dependency
+     * @since 2.20.2
+     */
+    String getClassifier();
+
+    /**
      * Returns the old version of the dependency
      * @return old version the dependency
      * @since 2.14.0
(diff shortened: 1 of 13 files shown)


ggkochanski:
@andrzejj0 please go ahead, you know the structure better then me 😅 

The classifier makes only sens for dependencies, so sth like this should be sufficient:
```
class DefaultDependencyVersionChange {
...
        public Builder withDependency(Dependency dependency) {
            groupId = dependency.getGroupId();
            artifactId = dependency.getArtifactId();
+            classifier = dependency.getClassifier();
            oldVersion = dependency.getVersion();
            return this;
        }

        public Builder withArtifact(Artifact artifact) {
            groupId = artifact.getGroupId();
            artifactId = artifact.getArtifactId();
  +          classifier = artifact.getClassifier();
            oldVersion = artifact.getVersion();
            return this;
        }
```

ggkochanski:
code format fixed, schema updated, integration test added

and waiting for events to unfold :blush:

PR #1340, opened 2026-01-26
set-property should add the property if it does not yet exist

### LINKED ISSUE #1268: set-property should add the property if it does not yet exist
`set-property` should add the respective property if it does not exist physically in the current pom.xml.

Context: I have a certain property with a default value in the parent pom for my project and I want to override the property in projects which are using that parent pom (which is located in a separate repo).

It might be a good idea to make this an opt in.

mandritschke:
Fixes #1268

* instead of just updating existing properties allows to insert ~completely new properties~ new properties overriding properties from parent POM
* works for properties on the project level as well as on the profile level (that is if a profile id is given)
* though this implementation won't create new profiles at the moment, this could be a possible enhancement, let me know if this sounds useful
* formatting was kind of a challenge

*Usage*
```
mvn org.codehaus.mojo:versions-maven-plugin:2.21.1-SNAPSHOT:set-property \
  -Dproperty=mockitoVersion \
  -DnewVersion=4.11.1 \
  -Dinsert
```

Notice that set-properties will be executed recursively on child modules if there are any.
If you don't want the property to be inserted into all of these child modules use the `-N` option.

### DIFF
--- a/versions-common/src/main/java/org/codehaus/mojo/versions/api/PomHelper.java
+++ b/versions-common/src/main/java/org/codehaus/mojo/versions/api/PomHelper.java
@@ -81,6 +81,7 @@ import org.apache.maven.project.ProjectBuildingRequest;
 import org.apache.maven.project.ProjectBuildingResult;
 import org.apache.maven.shared.utils.io.IOUtil;
 import org.codehaus.mojo.versions.rewriting.MutableXMLStreamReader;
+import org.codehaus.mojo.versions.rewriting.PropertyVersionInternal;
 import org.codehaus.mojo.versions.utils.ArtifactFactory;
 import org.codehaus.mojo.versions.utils.ModelNode;
 import org.codehaus.mojo.versions.utils.RegexUtils;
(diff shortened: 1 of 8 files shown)


wilx:
Thanks for the contribution. I found two regressions and one gap:

- [Properties before a profile’s `<id>`](https://github.com/mandritschke/versions/blob/81d1e309f71653ffaf10c7611f0edff91ab760e1/versions-common/src/main/java/org/codehaus/mojo/versions/rewriting/PropertyVersionInternal.java#L272) no longer update; with `-Dinsert`, this creates a duplicate `<properties>` block that Maven rejects.
- [Stopping at the first matching property](https://github.com/mandritschke/versions/blob/81d1e309f71653ffaf10c7611f0edff91ab760e1/versions-common/src/main/java/org/codehaus/mojo/versions/rewriting/PropertyVersionInternal.java#L377) regresses repeated property names: Maven uses the last declaration, which the previous implementation updated.
- [The existing null-value guard](https://github.com/mandritschke/versions/blob/81d1e309f71653ffaf10c7611f0edff91ab760e1/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetPropertyMojo.java#L208) prevents `-Dinsert -DprofileId=...` from inserting a project/inherited property into a profile lacking a local declaration.

Reproduced all three; 80 selected unit tests pass.

PR #1360, opened 2026-04-11
fix(set): ensure parent POM is updated when goal is executed from a c…

MohammedGhallab:
### 📝 Description
This PR addresses a regression introduced in version **2.21.0** where the `versions:set` goal failed to update the parent POM and shared references (like BOMs) when executed from within a child module or when using the `-pl` (project list) flag.

---

### 🔍 Technical Analysis
In the **2.21.0** release, the plugin's scope was overly restricted. In a **Monorepo** or **Multi-module** structure:
* **Inheritance Issue:** The plugin would throw `Project version is inherited from parent` because it couldn't contextually reach the parent to apply the change.
* **Broken References:** Even if the child version was updated, the references in a central **BOM** or other sibling modules remained unchanged, breaking the build integrity.

---

### 🛠️ The Solution
I have reintroduced **Cross-module update propagation** by ensuring the `localRoot` is explicitly included in the update cycle.

#### **Key Changes in `SetMojo.java`:**
* **Root Discovery:** Utilized `getLocalRoot(project)` to identify the physical aggregator root in the local file system.
* **File Persistence:** Added the `localRoot` POM file to the `files` set. This ensures that `PomHelper` has the necessary file handle to persist changes to the parent and its managed dependencies.
* **Safety First:** Added null-checks for `localRoot.getFile()` to prevent `NullPointerException` in virtual environments.

---

### 🧪 Verification & Testing
* **Unit Tests:** All **150+** existing tests passed successfully (`mvn test`).
* **Manual Regression Test:**
  1. Created a multi-module project with a Parent, BOM, and Child.
  2. Executed: `mvn -pl child-service versions:set -DnewVersion=1.2.3`
  3. **Result:** ✅ Parent, Child, and BOM were all updated correctly.
* **Environment:** Tested on Java 17 and Maven 3.9.x.

---

### 🎯 Impact
* Restores compatibility for **Monorepo** workflows.
* Ensures data consistency across the entire Maven Reactor.
* Fixes the `BUILD FAILURE` reported by users transitioning from **2.20.1**.

### DIFF
--- a/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
+++ b/versions-maven-plugin/src/main/java/org/codehaus/mojo/versions/SetMojo.java
@@ -65,7 +65,8 @@ import org.eclipse.aether.RepositorySystem;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 
 /**
- * Sets the current project's version and based on that change propagates that change onto any child modules as
+ * Sets the current project's version and based on that change propagates that
+ * change onto any child modules as
  * necessary.
  *
  * @author Stephen Connolly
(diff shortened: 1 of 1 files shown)


slachiewicz:
please rebase and fix build

PR #1394, opened 2026-08-31
Apply excludes/includes to candidate versions, not just the current one

### LINKED ISSUE #1378: Regression in 2.19.0: use-latest-releases no longer applies excludes to candidate versions
### Description

Starting with Versions Maven Plugin 2.19.0, `use-latest-releases` no longer applies version-specific `excludes` patterns to candidate versions.

This worked in 2.18.0 but fails in 2.19.0 and 2.21.0.

### Minimal POM

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0">

  <groupId>example</groupId>
  <artifactId>versions-plugin-reproducer</artifactId>
  <version>1.0.0</version>

  <parent>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-parent</artifactId>
    <version>3.5.0</version>
    <relativePath/>
  </parent>
</project>
```

### Reproduction

Using the documented five-part exclude format
`groupId:artifactId:type:classifier:version:`

```sh
mvn org.codehaus.mojo:versions-maven-plugin:2.18.0:use-latest-releases \
  -DprocessParent=true \
  -DprocessDependencies=false \
  -DprocessDependencyManagement=false \
  -DallowMajorUpdates=true \
  -DgenerateBackupPoms=false \
  '-Dexcludes=org.springframework.boot:spring-boot-starter-parent:pom:*:4*'
```

Version `2.18.0` does not select a Spring Boot 4 version.

Run the same command with 2.21.0:

```sh
mvn org.codehaus.mojo:versions-maven-plugin:2.21.0:use-latest-releases \
  -DprocessParent=true \
  -DprocessDependencies=false \
  -DprocessDependencyManagement=false \
  -DallowMajorUpdates=true \
  -DgenerateBackupPoms=false \
  '-Dexcludes=org.springframework.boot:spring-boot-starter-parent:pom:*:4*'
```

The parent is upgraded to a Spring Boot 4 version despite the exclude.
The regression can also be reproduced with the previously supported four-part pattern:
`org.springframework.boot:spring-boot-starter-parent:pom:4*`

### Expected behavior

Candidate versions matching the version component of excludes should not be considered. The newest eligible Spring Boot 3 version should be selected.

### Actual behavior

The current artifact is checked against excludes, but candidate versions are not. Because the current version starts with 3, the exclusion for 4* does not match, and a Spring Boot 4 version is selected.

### Suspected cause

This appears to have been introduced by:

- PR #1202
- Commit: 74624c5d8c435a8f07c24cf1cb832a134ab144f1
- First released in 2.19.0

Before that change, `UseLatestReleasesMojo.getLastFiltered()` created an artifact for each candidate version and applied `isIncluded()` to it.
PR #1202 removed `getLastFiltered().` The shared implementation now calls `isIncluded()` only for the artifact containing the current version, before selecting the newest candidate.

Could candidate-version filtering be restored for use-latest-releases?

adityaanikam:
Fixes #1378

## Problem

`getUpdates()` in `UseLatestVersionsMojoBase` calls `isIncluded()` exactly once per dependency, against the artifact's *current* version, before fetching the list of newer candidate versions:

```java
Artifact artifact = toArtifact(dep);
if (!isIncluded(artifact)) {
    return;
}
...
ArtifactVersions versions = getHelper().lookupArtifactVersions(artifact, false);
versionProducer(Arrays.stream(versions.getNewerVersions(...))
                .filter(this::artifactVersionsFilter))
        .ifPresent(ver -> updates.add(...));
```

A version-specific exclude pattern (e.g. `groupId:artifactId:*:*:4*`) only ever gets checked against the *current* version. If the current version doesn't match the pattern, every candidate version is added unfiltered — including the ones the pattern was meant to reject. In the reported case, an exclude targeting Spring Boot `4*` had no effect because the current version was `3.x`, and the newest `4.x` candidate got selected anyway.

This is a regression from #1202, which removed the old `UseLatestReleasesMojo.getLastFiltered()` method. That method used to build an `Artifact` per candidate version and run each one through `isIncluded()`. The shared `getUpdates()` path in `UseLatestVersionsMojoBase` (used by both `UseLatestReleasesMojo` and `UseLatestVersionsMojo`) never got an equivalent replacement.

## Fix

Restore the per-candidate check: after filtering the candidate stream with `artifactVersionsFilter`, build an `Artifact` for each candidate (same groupId/artifactId/type/classifier/scope as the current dependency, but with the candidate's version) and run it through `isIncluded()` too.

## Testing

Added `UseLatestReleasesMojoTest#testExcludeFilterAppliesToCandidateVersion`, which reproduces the exact shape of the report: a dependency whose current version doesn't match the exclude pattern, but whose only available newer release does. Verified with a negative control — reverting only the fix (keeping the new test) reproduces the failure precisely (the excluded candidate version still gets selected); restoring the fix passes. Full `UseLatestReleasesMojoTest` suite (8 tests) passes, including the existing `testIncludeFilter`/`testExcludeFilter` coverage, which only ever exercised whole-artifact excludes and so didn't catch this regression.


versions-maven-plugin/src/test/java/org/codehaus/mojo/versions/UseLatestReleasesMojoTest.java
```diff
@@ -133,4 +135,38 @@ public void testDisallowDowngrade()
...
+                                .build()));
+        setVariableValueToObject(mojo, "processDependencies", true);
+        // The current version (1.0) does not match this pattern, but the only
+        // available newer release (2.0) does -- gh-1378.
+        setVariableValueToObject(mojo, "excludes", new String[] {"default-group:other-artifact:*:*:2*"});
+
+        tryUpdate();
+        assertThat(
+                changeRecorder.getChanges(),
+                not(hasItem(new DefaultDependencyVersionChange("default-group", "other-artifact", "1.0", "2.0"))));
```
wilx, line 170: Please include the fixture's `"default"` classifier in the expected change. The four-argument constructor supplies a `null` classifier, while current master (`99bbbb9c`, following `b9436791`) records the dependency's classifier and includes it in `DefaultDependencyVersionChange.equals()`. Consequently, after merging onto current master, this `not(hasItem(...))` assertion passes even when the excluded `1.0 -> 2.0` update is recorded.

I verified that the negative control fails correctly on this PR's original base, but passes incorrectly on current master with only the new test applied. Changing the expectation to `new DefaultDependencyVersionChange("default-group", "other-artifact", "default", "1.0", "2.0")` makes the negative control fail again; restoring the production fix then passes all 40 focused tests.

wilx (commented):
The production fix resolves #1378 with both exclusion formats. One regression-test assertion needs updating for current master.

adityaanikam:
Good catch, thanks. Confirmed: the four-argument constructor delegates with a null classifier, the fixture builds other-artifact with classifier `default`, and equals() compares it, so that assertion passed regardless of what the production code did. Updated to the five-argument form and merged current master. Negative control re-run on this branch: with UseLatestVersionsMojoBase reverted the test fails on that assertion, and passes again once it is restored.

